PR #22, opened 2015-12-02
Make method name Section.HasKey match documentation.

jds:
This commit changes `Section.Haskey` to `Section.HasKey` (notice the capitalization on the k), which is what the documentation states (and which reflects the naming convention for this library). There is also a very minor documentation fix for `Section.HasValue`.


Unknwon:
How could I possibly miss that... thank you! :+1:

Issue #23, opened 2015-12-03
Fails if sections have comments on the same line.


Timmmm:
For example

    [a_section] ;This is my favourite section

It fails because of this code:

    	case line[0] == '[' && line[length-1] == ']': // New sction.

It only checks that the first and list characters are [ and ]. If there is a comment then this isn't true.

Unknwon:
Fixed by #24

PR #24, opened 2015-12-03
Better comments support

Timmmm:
This adds support for comments on the end of section headers, e.g.

    [a_section] ; A comment

It also fixes a couple of places where only `#` comments were supports - not `;` comments.

Test cases are included (they pass).

This closes bug #23


Unknwon:
Thank you!

PR #28, opened 2015-12-20
Update ini.go

woniu369:
让strconv.Parse*,根据程序位数自动转换

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -502,7 +502,7 @@ func (k *Key) Uints(delim string) []uint {
 	strs := k.Strings(delim)
 	vals := make([]uint, len(strs))
 	for i := range strs {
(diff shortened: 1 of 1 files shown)


ini.go
```diff
@@ -502,7 +502,7 @@ func (k *Key) Uints(delim string) []uint {
 	strs := k.Strings(delim)
 	vals := make([]uint, len(strs))
 	for i := range strs {
-		u, _ := strconv.ParseUint(strs[i], 10, 64)
+		u, _ := strconv.ParseUint(strs[i], 10, 0)
```
Unknwon: Other than this, other places should use 64.

Unknwon:
Thanks!

PR #31, opened 2016-01-24
Exclude not parsed values from lists.

lyobzik:
I found that methods that return lists (Float64s, Ints, Int64s and etc.) do not handle parsing errors. For example, following code:
```go
config, _ := Load([]byte(_CONF_DATA))
stringKey := config.Section("array").Key("STRINGS")
fmt.Printf("string values: %v\n", stringKey.Strings(","))
fmt.Printf("float values: %v\n", stringKey.Float64s(","))
```

gives such result:
```
string values: [en zh de]
float values: [0 0 0]
```

I think that the best solution would be to return error from these methods, but it will break down back compatibility of interface.
Therefore I propose do not append to list such values, that was parsed with error.

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -535,6 +535,88 @@ func (k *Key) Times(delim string) []time.Time {
 	return k.TimesFormat(time.RFC3339, delim)
 }
 
+// ValidFloat64s returns list of float64 divided by given delimiter. If some value is not float, then
+// it will not be included to result list.
+func (k *Key) ValidFloat64s(delim string) []float64 {
+	strs := k.Strings(delim)
+	vals := make([]float64, 0, len(strs))
+	for i := range strs {
+		if val, err := strconv.ParseFloat(strs[i], 64); err == nil {
+			vals = append(vals, val)
+		}
+	}
+	return vals
+}
+
+// ValidInts returns list of int divided by given delimiter. If some value is not integer, then it will
+// not be included to result list.
+func (k *Key) ValidInts(delim string) []int {
+	strs := k.Strings(delim)
+	vals := make([]int, 0, len(strs))
+	for i := range strs {
+		if val, err := strconv.Atoi(strs[i]); err == nil {
+			vals = append(vals, val)
+		}
+	}
+	return vals
+}
+
+// ValidInt64s returns list of int64 divided by given delimiter. If some value is not 64-bit integer,
+// then it will not be included to result list.
+func (k *Key) ValidInt64s(delim string) []int64 {
+	strs := k.Strings(delim)
+	vals := make([]int64, 0, len(strs))
+	for i := range strs {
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks your PR!

These kind of methods are meant to ignore invalid input and treat as zero value in Go (`0.0` for `float`, empty string for `string`).

I suggest you create new methods like `ValidStrings` `ValidFloat64s`? Or any suggestions of name.

lyobzik:
Excuse me for long delay with answering.
I reverted my changes from original list methods and added `Valid` methods.

May I also add `Check` methods, that will return error on first invalid value, and extract common code from all list methods to helpers? I done this in separate commit.

Unknwon:
Thanks for following up!

> May I also add Check methods, that will return error on first invalid value, and extract common code from all list methods to helpers?

So what `Check` does is **return valid values** or **only return error**?

lyobzik:
I had not thought about this previously. But now i think `Check` methods must only return error. I amended my last commit as follows:
```
                if err != nil && returnOnInvalid {
-                       return vals, err
+                       return nil, err
                }
```
Also i added in tests check, that `Check` methods will return empty value on error.

Unknwon:
Got it. In this case I think `Check` might not the correct work? Maybe call these methods `Strictxxx`?

lyobzik:
I amended last commit and rename `Check` methods to `Strict`.
But i don't understand why these methods may work incorrect?

Unknwon:
> But i don't understand why these methods may work incorrect?

Incorrect?

lyobzik:
Erm... I don't understand this your question:
> In this case I think `Check` might not the correct work?

Do you mean that these methods may work incorrect in some cases or something else?
_Excuse me, my English is bad._

Unknwon:
Oops.... may bad... I was really trying to say `word`... :sob:

Unknwon:
Can you run `go test -cover -race` and post results here?

lyobzik:
```
 go test -cover -race
.
1 total assertion

...
4 total assertions

.............
17 total assertions

...................................................................................................................................................................................................................................................................
276 total assertions

................
292 total assertions

......
298 total assertions

..
300 total assertions

.....
305 total assertions


305 total assertions

....
309 total assertions

....................................
345 total assertions

..
347 total assertions

.....
352 total assertions

PASS
coverage: 96.8% of statements
ok  	_/home/leto/work/go/src/github.com/lyobzik/ini	1.068s
```

lyobzik:
And i am using go 1.5.3:
```
 go version
go version go1.5.3 linux/amd64
```

Unknwon:
Awesome! Thanks!

lyobzik:
Thank you!

Unknwon:
Added docs to README: https://github.com/go-ini/ini#auto-split-values-into-a-slice

PR #34, opened 2016-02-22
Add LooseLoad - fix #33

### LINKED ISSUE #33: Panic when submitting non-existing files to Load()
Hi,

I'm submitting a slice of three files to `ini.Load()`:

```go
cfg, err := ini.Load("/etc/securepass.conf", "/usr/local/etc/securepass.conf", "securepass.conf")
```

Only the latter exists and has the correct permissions set. Likewise the Python `ConfigParser` module, it would be great if the function could read the configuration from only those accessible or return an empty configuration if no existing files were provided.

```console
$ ~/go/src/github.com/garlsecurity/go-securepass$ go run spctl/main.go 
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x4811b2]

goroutine 1 [running]:
gopkg.in/ini%2ev1.(*File).GetSection(0x0, 0x7fa130, 0x7, 0x0, 0x0, 0x0)
	/home/SMARTODDS/tregliaa/go/src/gopkg.in/ini.v1/ini.go:976 +0x72
main.loadConfiguration()
	/home/SMARTODDS/tregliaa/go/src/github.com/garlsecurity/go-securepass/spctl/main.go:50 +0xc2
main.main()
	/home/SMARTODDS/tregliaa/go/src/github.com/garlsecurity/go-securepass/spctl/main.go:60 +0x1c

goroutine 17 [syscall, locked to thread]:
runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:1721 +0x1
exit status 2
```

Thanks for considering.

alessio:
A common practice in applications development is to allow the user
to customise/extend program's settings by overriding global defaults
set in system-wide config files (typically installed under /etc)
with site-specific copies (e.g. /usr/local/etc) and user-specific
files generally stored in the user's home directory. It must be
noticed that this cascade approach allows the user to specify only
those options that he wishes to override, whilst the other values
will keep their value already set in the other configuration
files. In general, when a program couldn't find any configuration
file rather than crashing it falls back to hardcoded default values.
    
These changes introduce a new function LooseLoad that allows the
caller to pass a slice of filenames without worrying about their
existence or not. LooseLoad will attempt to load them in order.
If a file doesn't just, it will just be ignored and the function
won't return any error. In case no files existed, an empty
configuration will be returned to the caller.

This functionality is inspired by the Python's standard library
configparser.RawConfigParser.read() function:

  https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.read

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -123,16 +123,20 @@ type File struct {
 	// To keep data in order.
 	sectionList []string
 
+	// whether the parser should check for files existence first
+	looseMode bool
+
 	NameMapper
 }
 
 // newFile initializes File object with given data sources.
-func newFile(dataSources []dataSource) *File {
+func newFile(dataSources []dataSource, looseMode bool) *File {
 	return &File{
 		BlockMode:   true,
 		dataSources: dataSources,
 		sections:    make(map[string]*Section),
 		sectionList: make([]string, 0, 10),
+		looseMode:   looseMode,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


Unknwon:
Thanks!

alessio:
:+1:

PR #38, opened 2016-03-21
Add support for explicitly writing a header in the DEFAULT section

### LINKED ISSUE #37: Explicit DEFAULT section when writing to file
Currently, when adding/modifying a key in the `DEFAULT` section, there is no way to explicitly write the section name. For example, if I had an existing file, config.ini:

```
[DEFAULT]
key1 = value1

[another_section]
another_key = another_value
```

If I modified `key1` like so:

```
cfg, _ := ini.Load("config.ini")
key := cfg.Section("DEFAULT").Key("key1")
key.SetValue("value2")
cfg.SaveTo("config.ini")
```

This would strip the explicit `[DEFAULT]` section name from the file, resulting in:

```
key1 = value2

[another_section]
another_key = another_value
```

It would be great to have the ability to specify including the explicit `DEFAULT` section header.

nextrevision:
Adds support for explicitly writing a header for the DEFAULT section in files. Disabled by default, and can be enabled by setting the following in the implementation of the library:

```
ini.DefaultHeader = true
```

For example:

```
package main

import (
	"github.com/go-ini/ini"
)

func main() {
	ini.DefaultHeader = true
	cfg, _ := ini.Load("config.ini")
	key := cfg.Section("DEFAULT").Key("key1")
	key.SetValue("value2")
	cfg.SaveTo("config.ini")
}
```

Closes #37

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -56,6 +56,9 @@ var (
 	// Indicate whether to align "=" sign with spaces to produce pretty output
 	// or reduce all possible spaces for compact format.
 	PrettyFormat = true
+
+	// Explicitly write DEFAULT section header
+	DefaultHeader = false
 )
 
 func init() {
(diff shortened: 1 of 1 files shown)


Unknwon:
Thanks!

PR #42, opened 2016-04-08
Fix calls for KeysHash() in README files

loshkovskyi:
It looks like a typo in README's. I also verified the correct call in `key_test.go`.


Unknwon:
Thanks!

PR #43, opened 2016-04-13
Allow escaping comments

andreynering:
Allows escaping `#` and `;` caracteres putting a backslash before (`\#` and `\;`).

https://en.wikipedia.org/wiki/INI_file#Escape_characters

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -75,11 +75,20 @@ func (p *parser) readUntil(delim byte) ([]byte, error) {
 }
 
 func cleanComment(in []byte) ([]byte, bool) {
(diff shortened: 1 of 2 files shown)


nibogd:
Why was this closed? :grin: It's a useful feature!

andreynering:
I after found that this package handle that differently

You can escape comments with backsticks (`)

nibogd:
Thanks, i've found that too after of a bit searching, but it's less intuitive :) Maybe that should be added to README.

andreynering:
> Maybe that should be added to README

Agree

Issue #32, opened 2016-01-25
Parser should respect quoted or escaped delimiter


marcopaganini:
If we have a field defined as a slice and using space as a delimiter, say:

```
ExtraArgs   []string `ini:"extra_args" delim:" "`
```

An entry in the ini file formatted as:
```
extra_args = aaa bbb ccc
```

Correctly generates a slice of three string elements, "aaa", "bbb", and "ccc".

However, things get complicated if we need one of the elements to contain a space. I tried the following constructs without success:

```
extra_args = aaa "bbb and space" ccc

extra_args = aaa bbb\ and\ space ccc

extra_args = """
aaa
bbb and space
ccc"""
```

And neither seems to work.

Ideally, the delimiter shouldn't be matched inside quoted strings of it it's quoted with a backslash.

Unknwon:
Thank your feedback!

It is indeed a problem... but why can you just use `,` :trollface:

marcopaganini:
Oh,  but then we'd have the same problem with the comma, no? :)
The fact is, unless we allow quoting or escaping, there will always be one
character we can't use (the delmiter).

Unknwon:
If you have `,`, then use `,,`, then `,,,`, and `,,,,`, etc...

PR #46, opened 2016-05-08
Match last ] in line when parsing Section header

ijc25:
I have a config.ini where section names can contain "]", which leads to parse
errors. While proper support for quoting would be best (issue #32) simply
matching the last "]" rather than the first helps in this case.


Unknwon:
Thanks!

PR #47, opened 2016-05-09
Parsing to struct now supports require and struct flags, and int pars…

silasray:
…ing doesn't drop 0 values on the floor.


silasray:
Going to merge to fork first...

PR #49, opened 2016-05-12
remove # as comment

nathanphan:
situation:

in ini file I have key like
currency_fortmat="$ #"

When parse, it will return only "$

I think there is a better way but this is my workaround please consider and improve it.

thanks


plutov:
Thanks. @Unknwon, I was waiting for this fix :)

Unknwon:
Thanks your PR!

However, remove an existing behavior is not acceptable at this point.

You can do:

```ini
currency_fortmat=`$ #`
```

PR #50, opened 2016-05-20
Folded tags together, fixed parsing bug

silasray:
`ini` and `iniFlags` now all live in `ini`
Partially fixed bug in time.Time field parsing.  Fixing fully will take a broader refactor, as the buggy assumption is built in to the parser at a pretty low level.  The fundamental problem is that `reflect.TypeOf(time.Now()).Kind() == reflect.Struct` (and other complex types as well), and the parser assumes that you can accurately differentiate types for parsing based on Kind().

### DIFF
--- a/README.md
+++ b/README.md
@@ -611,7 +611,7 @@ Age = 21
 ```
 
 ### MapTo special flags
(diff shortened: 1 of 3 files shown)


Unknwon:
What happened?

silasray:
Accidentally PRed this to here instead of our repo.  This change should be in #51.

PR #51, opened 2016-05-20
Bug fixes and improvements

silasray:
There are a number of bug fixes and improvements made here.  If there is an area where things aren't covered, it's the write out to ini side of things, but all the tests are passing.  These changes are pretty much exclusively to do with struct-based parsing.  Feature additions are noted in the README as well.
- Struct parsing now supports *struct typed fields
- New tags to cause the parser to error if fields don't parse or don't exist
- Tweak to empty field parsing so that explicitly set 0 values in the input data will actually overwrite non-zero defaults in the target struct
- Started to fix issues arising from incorrect assumptions in parser about Kind() of time.Time

Sorry for the false start PRs earlier.  Please let me know what's needed to get this merged.  Thanks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,12 @@
+.PHONY: build test bench vet
+
+build: vet bench
+
+test:
+	go test -v -cover -race
+
+bench:
+	go test -v -cover -race -test.bench=. -test.benchmem
+
+vet:
+	go vet
--- a/README.md
+++ b/README.md
@@ -62,6 +62,14 @@ When you cannot decide how many data sources to load at the beginning, you still
 err := cfg.Append("other file", []byte("other raw data"))
 ```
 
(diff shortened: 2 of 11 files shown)


Unknwon:
Thanks your PR!

But I would recommend sent each improvement/bugfix separately.

silasray:
Yeah, I was making these changes as I needed them for my project.  I'll see what I can do to split them up...

silasray:
@Unknwon splitting these up is going to be non-trivial, I think.  Could you look at the changes and tell me  which you'd be interested in taking, so I won't spend time on things you aren't interested in taking?  I added explanation for all the new user-facing features to the docs, if that's a help.

Unknwon:
Hmm, we have conflict now..

silasray:
Yeah, the bug fix did for the ticket I filed created the conflict.  I can base off that when I break these features up in to new pull requests, but I just want to know which features you are interested in.

Unknwon:
Hmm... I can't really distinguish features in current file change diff right now, all of them look interesting.

PR #52, opened 2016-06-01
Retrieve parent keys available to a child section

AT1012:
This PR lets us retrieve a list of parent keys that are
available to the child section.

### DIFF
--- a/key_test.go
+++ b/key_test.go
@@ -59,6 +59,13 @@ func Test_Key(t *testing.T) {
 			}
 		})
 
+		Convey("Get parent-keys that are available to the child section", func() {
+			parentKeys := cfg.Section("package.sub").ParentKeys()
+			for _, k := range parentKeys {
+				So(k.Name(), ShouldEqual, "CLONE_URL")
+			}
+		})
+
 		Convey("Get overwrite value", func() {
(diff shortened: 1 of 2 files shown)


key_test.go
```diff
@@ -59,6 +59,13 @@ func Test_Key(t *testing.T) {
 			}
 		})
 
+		Convey("Get parent-keys that are available to the child section", func() {
+			parentKeys := cfg.Section("package.sub").ParentKeys()
+			for _, k := range parentKeys {
+				So(k.Name(), ShouldEqual, fmt.Sprintf("%s", "CLONE_URL"))
```
Unknwon: I think you can replace `fmt.Sprintf("%s", "CLONE_URL")` with `"CLONE_URL"`, plain text string does not need format.

Unknwon:
Other than that, LGTM.

key_test.go
AT1012: Fixed.

Unknwon:
Thanks, will tag the version ASAP.

AT1012:
Thank you!

Unknwon:
https://github.com/go-ini/ini/releases/tag/v1.12.0

PR #55, opened 2016-06-25
Fix panic from incorrect alignment length calculation

ryaka:
The alignment length calculation did not take into account that
keys may be added to if they contained "\"", "=", ":", or "`".

This gave an under calculation and resulted in index out of
bounds when this happened to be the longest key.

Please feel free to give any feedback or issues with the following request


ryaka:
Please feel free to give any feedback or issues with the following request

Unknwon:
Thanks!

PR #60, opened 2016-08-03
Add support for omitempty tag option

### LINKED ISSUE #59: Allow omitting empty keys
The default golang JSON (un)marshal implementation respects `omitempty` inside a field's tag which will effectively omit such field from the resulting JSON if it's not defined in a given struct.

```go
Field int `json:"myName,omitempty"`
```

I was wondering if there's any interest in such feature for this library - i.e. support something like this:

```go
type Profile struct {
	First  string `ini:"first"`
	Second string `ini:"second,omitempty"`
}
```

The current implementation doesn't interpret commas in any special way which results in the following output:

```ini
[my-section]
first            = custom-value
second,omitempty = 
```

The expected behaviour would be

```ini
[my-section]
first = custom-value
```

Is there any reason to support commas inside key names? 😺

radeksimko:
Closes #59

The `isEmptyValue` is shamelessly stolen from https://github.com/golang/go/tree/master/src/encoding/json , I just added support for `time.Time`. 😉

### DIFF
--- a/struct.go
+++ b/struct.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 	"unicode"
 )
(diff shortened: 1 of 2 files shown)


struct.go
```diff
@@ -334,11 +355,23 @@ func (s *Section) reflectFrom(val reflect.Value) error {
 		tpField := typ.Field(i)
 
 		tag := tpField.Tag.Get("ini")
-		if tag == "-" {
+		opts := strings.Split(tag, ",")
+		if len(opts) > 2 {
+			return fmt.Errorf("Expected max. 2 comma-separated values in a tag, %d given", len(opts))
+		}
+
+		if opts[0] == "-" {
```
Unknwon: If tag is `-`, then should get out of loop directly, so we need `if tag == "-" {`, then do the `omitempty` check.

radeksimko: PR updated.

Unknwon:
Thanks!

PR #65, opened 2016-08-26
Support omitempty in tags during MapTo

jbowes:
omitempty must be ignored when unmarshalling into a struct, so that the
key name is parsed correctly.


Unknwon:
Thanks!

PR #68, opened 2016-10-31
 Fix example code about "Section" method in README.

PuHsiu:
It's more clearly to describe not only the default section but also the "section name" can be to use in this method.

Changes to be committed:
	modified:   README.md
	modified:   README_ZH.md


PuHsiu:
I'm sorry that mix two type of commit into one. :(

PR #72, opened 2016-11-26
Added support for unparseable sections.

gavbaa:
There are public standards ( example: https://github.com/ADL-AICC/AICC-Document-Archive/releases/tag/cmi001v4 ) that are based around INI formats that allow for both standard key/value sections as well as sections that contain raw data that cannot be parsed.  My attempt here is to provide a mechanism for leaving unparsed sections in their raw form, while otherwise meeting the INI parsing standards.  It's off by default and can only be opted-into on a per-section basis by section name as part of the LoadOptions config.  Tests included, and also cover the case-insensitive section names.


Unknwon:
Thanks!

PR #100, opened 2017-04-26
Config pointer does not have GetKey directly

hhercules:
Section has GetKey function, and not config pointer directly.

```
cmd/export.go:32: cfg.GetKey undefined (type *ini.File has no field or method GetKey)
```


hhercules:
@Unknwon The build failed on master but does not seem to be related to to ReadMe update.

Unknwon:
Thank you!

hhercules:
👍

PR #110, opened 2017-07-02
Add support for double quote on item value

lunny:
This PR will support the format which is used on crowdin download file.

```ini
create_repo="创建了仓库 <a href=\"%s\">%s</a>"
```

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -202,6 +202,8 @@ func (p *parser) readValue(in []byte, ignoreContinuation, ignoreInlineComment bo
 	var valQuote string
 	if len(line) > 3 && string(line[0:3]) == `"""` {
 		valQuote = `"""`
+	} else if line[0] == '"' {
+		valQuote = `"`
 	} else if line[0] == '`' {
 		valQuote = "`"
 	}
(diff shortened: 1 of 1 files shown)


parser.go
```diff
@@ -202,6 +202,8 @@ func (p *parser) readValue(in []byte, ignoreContinuation, ignoreInlineComment bo
 	var valQuote string
 	if len(line) > 3 && string(line[0:3]) == `"""` {
 		valQuote = `"""`
+	} else if line[0] == '"' {
+		valQuote = `"`
```
Unknwon: According to https://github.com/go-ini/ini#working-with-values:

image

So this won't be supported as default behavior.

parser.go
```diff
@@ -213,6 +215,9 @@ func (p *parser) readValue(in []byte, ignoreContinuation, ignoreInlineComment bo
 		if pos == -1 {
 			return p.readMultilines(line, line[startIdx:], valQuote)
 		}
+		if valQuote == `"` {
+			return strings.Replace(line[startIdx:pos+startIdx], `\"`, `"`, -1), nil
```
Unknwon: How do you distinguish user actually wants `\"` instead of replacing them to `"`"?

Unknwon:
You can now use this feature via:

```go
cfg, err := ini.LoadSources(ini.LoadOptions{UnescapeValueDoubleQuotes: true}, ...))
```

PR #115, opened 2017-08-04
Avoid using temp file during write

ethantkoenig:
Some slight refactoring to avoid writing/renaming a temp file when writing a ini file. Using a temp file is problematic if the file you are writing to is mounted inside a Docker container (see https://github.com/go-gitea/gitea/issues/2197).

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -24,10 +24,8 @@ import (
 	"os"
 	"regexp"
 	"runtime"
-	"strconv"
 	"strings"
 	"sync"
-	"time"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


ethantkoenig:
CI build passed for all Go versions 1.5+, failed for Go 1.4 with the following:
```
../../smartystreets/logging/logging.go:61: this.Logger.SetOutput undefined (type *log.Logger has no field or method SetOutput)
../../smartystreets/logging/logging.go:68: undefined: log.Output
```

Unknwon:
Thanks your PR!

I think the problem is Go 1.4, I saw goconvey supports from 1.2.x. Just a small change on CI config file, I'll do that after merge this PR.

Unknwon:
Hmm... CI is still failing. Anyway I'll remove Go 1.4 from CI file.

PR #116, opened 2017-08-27
Escape delimiters in string slice values

### LINKED ISSUE #32: Parser should respect quoted or escaped delimiter

asciimoo:
Use backslash to escape delimiters.

Fixes #32

### DIFF
--- a/ini_test.go
+++ b/ini_test.go
@@ -87,6 +87,13 @@ key  = "value"
 key2 = "value2" ; This is a comment for key2
 key3 = "one", "two", "three"
 
+[string escapes]
+key1 = value1, value2, value3
+key2 = value1\, value2
+key3 = val\ue1, value2
+key4 = value1\\, value\\\\2
+key5 = value1\,, value2
+
 [advance]
 value with quotes = "some value"
 value quote2 again = 'some value'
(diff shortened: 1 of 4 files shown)


Unknwon:
Thanks for the PR!

But the test cases didn't include the one proposed in https://github.com/go-ini/ini/issues/32, can you test on that and make sure it works?

asciimoo:
> But the test cases didn't include the one proposed in #32, can you test on that and make sure it works?

Which test case isn't included?

Unknwon:
Use space as delimiter and escape some spaces:

image

asciimoo:
@Unknwon added in 840d413

PR #119, opened 2017-09-15
map slices in strict mode too

peter-vaczi:
Hi,

If I map a section to a struct with StrictMapTo instead of MapTo then fields with slice type are not mapped due to an 'if isStrict { return err}' construct in setSliceWithProperType. 
The PR is correcting this.

-ptr-


Unknwon:
Thank you!

PR #126, opened 2017-10-24
Fix integer overflow

gustavosbarreto:

time.Duration conversion in 32 bits systems. If duration is less than 16m50s the conversion will fail.

Yes


Unknwon:
Thank you!

PR #130, opened 2017-11-13
Keys should not consider themself for transformation.

apauzies:

NAME = ini
[package]
NAME = $(NAME)

I believe NAME in section [package] should evaluate to ini

nope :/


Unknwon:
Thank you!

Unknwon:
I have added tests for you.

apauzies:
Great! Thanks!

PR #133, opened 2018-02-25
remove trailing whitespace and add missing newline

carterjones:

Trailing whitespaces are present, which causes issues when people use a combination of vendored go code and pre-commit hooks that trim whitespace. Every time we update our vendored code that includes this (fantastic!) library, it pulls down the whitespace. Then, our pre-commit removes the whitespace and adds in the missing newline.

No. This does not affect the functionality of the library.


Unknwon:
Thank you!

Unknwon:
Just a note, the changes are not tagged. You'll be fine if your app/tool is pulling from `github.com/go-ini/ini`. Otherwise, `gopkg.in/ini.v1` has not yet included this change until I tag a new version (let me know if you need a new tag).

carterjones:
No tag is necessary. I'm pulling from `github.com/go-ini/ini`. Thanks for all your work on this project!

PR #143, opened 2018-04-10
Add Support for spaces around '=' w/o PrettyFormat

zerotao:
The default git config format is equivalent to PrettyFormat w/o
alignment (i.e. just spaces around '=') and SaveToIndent(..., "\t")

This patch allows for setting PrettyFormat=false and PrettySpaces=True
to achieve the same output as the git config default.

It is backwards compatible in that current use case output will not be
affected unless you specifically disable PrettyFormat AND enable
PrettySpaces.


Unknwon:
Thank you!

PR #146, opened 2018-05-23
Fix inline value parsing

syndbg:

Edge case has a value that contains an inline symbol such as `;` or `#`.

Currently parsing a value such as:

```
value = val#ue
```

would result into just `val` if `IgnoreInlineComment` is false.

If `IgnoreInlineComment` is true it would result into the full value, but it's incorrect to depend on this.

With the fix value would be parsed as `val#ue`.

Yes

@Unknwon

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -145,6 +145,11 @@ type LoadOptions struct {
 	// Relevant quote:  Values can also span multiple lines, as long as they are indented deeper
 	// than the first line of the value.
 	AllowPythonMultilineValues bool
+	// SpaceBeforeInlineComment indicates whether to allow comment symbols (\# and \;) inside value.
+	// Docs: https://docs.python.org/2/library/configparser.html
+	// Quote: Comments may appear on their own in an otherwise empty line, or may be entered in lines holding values or section names.
+	// In the latter case, they need to be preceded by a whitespace character to be recognized as a comment.
+	SpaceBeforeInlineComment bool
 	// UnescapeValueDoubleQuotes indicates whether to unescape double quotes inside value to regular format
 	// when value is surrounded by double quotes, e.g. key="a \"value\"" => key=a "value"
 	UnescapeValueDoubleQuotes bool
--- a/ini_test.go
+++ b/ini_test.go
@@ -65,7 +65,7 @@ NAME = Unknwon
 		So(err, ShouldBeNil)
 		So(f, ShouldNotBeNil)
 
-		// Vaildate values make sure all sources are loaded correctly
+		// Validate values make sure all sources are loaded correctly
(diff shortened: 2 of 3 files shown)


Unknwon:
Hi, comment isn't necessary comes with a space before the comment symbol (e.g. `#` or `;`), I don't think this assumption is correct.

syndbg:
@Unknwon Unfortunately, I can't reference a source-of-truth INI file specification since it varies greatly between implementations.

What I can reference is various implementations:

* `Comments may appear on their own in an otherwise empty line, or may be entered in lines holding values or section names. In the latter case, they need to be preceded by a whitespace character to be recognized as a comment. (For backwards compatibility, only ; starts an inline comment, while # does not.)`  https://docs.python.org/2/library/configparser.html
* `Any lines that begin with a comment character (;), after any whitespace, are ignored` http://www.nongnu.org/chmspec/latest/INI.html

A win-win solution is to hide it behind a feature flag. I don't really know how to name it, though since it's hard to pin-point the functionality to a specific platform or semantic of INI files.

Perhaps `SpaceBeforeInlineComment` ?

Unknwon:
I think this is overkill, I believe `IgnoreInlineComment` is sufficient enough, not to say you can use backquotes or three double quotes to reserve any string.

syndbg:
@Unknwon For my use case it's definitely not. It's multiple INI files provided by external parties that contain generated values (that also include `#` and/or `;`) and also inline comments.

Unknwon:
OK, make sense. `SpaceBeforeInlineComment` sounds good to me.

syndbg:
@Unknwon Sorry for the delay, was national holiday yesterday. 

I've updated, added `SpaceBeforeInlineComment` and tests

ini.go
```diff
@@ -145,6 +145,12 @@ type LoadOptions struct {
 	// Relevant quote:  Values can also span multiple lines, as long as they are indented deeper
 	// than the first line of the value.
 	AllowPythonMultilineValues bool
+	// SpaceBeforeInlineComment indicates whether to allow comment symbols (\# and \;) inside value.
+	// Docs: https://docs.python.org/2/library/configparser.html
+	// Quote: Comments may appear on their own in an otherwise empty line, or may be entered in lines holding values or section names.
+	// In the latter case, they need to be preceded by a whitespace character to be recognized as a comment.
+	// (For backwards compatibility, only ; starts an inline comment, while # does not.)
```
Unknwon: Is this line unnecessary? I see both ` #` and ` ;` added in the PR.

syndbg: Nope. Removed :+1:

Unknwon:
Thank you!

PR #148, opened 2018-07-06
Support *struct type in struct field.

ideadawn:
```ini
[a]
s = "test"
```

```go
type A struct {
  S string `ini:"s"`
}
type B struct {
  A *A `ini:a`
}
```

before: It's an error returned.
after: Successfully parsed the ini file.

### DIFF
--- a/struct.go
+++ b/struct.go
@@ -244,7 +244,7 @@ func (s *Section) mapTo(val reflect.Value, isStrict bool) error {
 			continue
 		}
 
-		isAnonymous := tpField.Type.Kind() == reflect.Ptr && tpField.Anonymous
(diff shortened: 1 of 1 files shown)


struct.go
```diff
@@ -244,7 +244,7 @@ func (s *Section) mapTo(val reflect.Value, isStrict bool) error {
 			continue
 		}
 
-		isAnonymous := tpField.Type.Kind() == reflect.Ptr && tpField.Anonymous
+		isAnonymous := tpField.Type.Kind() == reflect.Ptr && (tpField.Anonymous || tpField.Type.Elem().Kind() == reflect.Struct)
```
Unknwon: The purpose of `isAnonymous` variable is to indicate if the field is anonymous, what you have edited might work at this time, but it is not what it meant to be.

You should consider modifying line 248.

PR #154, opened 2018-07-22
Fix doc for ShadowLoad

Pixdigit:
The doc for ShadowLoad says InsenstiveLoad in its description.

Cannot test for documentation


Unknwon:
Thank you!

Issue #161, opened 2018-09-17
Add option to have padding spaces around equal sign when PrettyFormat is disabled


haroldHT:
There should be a way to close "PrettyFormat" when you init config
I do not want to " a         = 123"
I just want to "a = 123"

haroldHT:
you can use `ini.PrettyFormat=false`
But it will be 
`a=1
ab=3`
What I want is 
`a = 1
ab = 3`

Can I choose the indent after key?

haroldHT:
By the way

haroldHT:
I want to support multiple the same sections like：
`[global]
a=1
[global]
ab=1
`

The reason  why I do that is I do not want to change some configuration format.

Unknwon:
Hi, multiple section with same name is not and will not be supported at least in current version of this package. I've adjusted the title for your initial feature request.

haroldHT:
我可以加入开发吗，请问

Unknwon:
Yes you can send a pull request :grin:

Unknwon:
This is claimed to be fixed by merging #164.

PR #164, opened 2018-09-29
Feature-#161-Add-DefaultFormat

haroldHT:
Add option to have padding spaces around equal sign when PrettyFormat is disabled #161

eg:

> ini.PrettyFormat = false
>         ini.DefaultFormatLeft = "  "
>         ini.DefaultFormatRight = " "

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -48,6 +48,10 @@ var (
 	// at package init time.
 	LineBreak = "\n"
 
+	// Place custom spaces when close PrettyFormat and PrettyEqual
+	DefaultFormatLeft = ""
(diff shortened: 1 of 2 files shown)


ini.go
```diff
@@ -48,6 +48,10 @@ var (
 	// at package init time.
 	LineBreak = "\n"
 
+	// Place custom spaces when close PrettyFormat and PrettyEqual
```
Unknwon: s/when close PrettyFormat and PrettyEqual/when PrettyFormat and PrettyEqual are both disabled/

Unknwon:
Thank you!

PR #177, opened 2018-12-21
Fix recursive char stripping when parsing recursive values

garnermccloud:
`strings.TrimRight(noption, ")s")` will strip strings that end in "s)s" which wasn't intended.  We should only string the first 2 characters and last two characters of the string since we know they will be "%(" and  ")s".

yes


Unknwon:
Thank you!

Issue #179, opened 2019-01-14
Add option to preserve double quotes


drakkan:
I have an ini file that contains this entry inside a section:

```
DefaultTimezone = "GMT+01 (Amsterdam, Berlin, Rome, Stockholm, Warsaw)"
```

if I read this ini file from go and I change some other values, not the DefaultTimezone, after saving the file I have now:

```
DefaultTimezone=GMT+01 (Amsterdam, Berlin, Rome, Stockholm, Warsaw)
```

This ini file is part of a Qt application, QSettings use double quotes as escape chars, so if I now read the ini from Qt after saving it from go I cannot read DefaultTimezone as string anymore. Qt ini parser convert to stringlist this setting since it contains commas and the double quotes were removed. 

I saw an old discussion about this in #25

what do you think about preserving quotes and/or add a setting to add double quotes to each value?

Unknwon:
Thanks your feedback!

>what do you think about preserving quotes and/or add a setting to add double quotes to each value?

Sounds like a workable solution.

PR #182, opened 2019-02-06
parser: add option to preserve surrounded quote

drakkan:
see #179 

what do you think about? It works fine for my use case, thanks


Unknwon:
Thanks! Will review ASAP.

Unknwon:
LGTM, merging.

PR #189, opened 2019-04-25
struct: map into fields that are struct pointers

vishen:

Allow 'mapTo' to load into fields that are struct pointers iff there is
a section for that field and fieldname. The field is initialised to
pointer type and then the data is loaded.

Yes.


Unknwon:
Thank you!

PR #194, opened 2019-05-22
support Python multi-line values with empty first line

shahms:
Multline line values whose initial line was empty were not parsed. This style of list is commonly used in a couple of places, but the main one I've found is LLVMBuild.txt (an ini-style build configuration used by LLVM).

Yes


Unknwon:
Thank you!

PR #200, opened 2019-08-01
struct: support for *bool

greut:

go-ini support `bool` but not `*bool` which is a clever way to work around the infamous `bool,omitempty` issue that Go serialization has. Then, a three state value exists.

```ini
Optional =
Optional = true
Optional = false
```

Needs for the editorconfig, https://github.com/editorconfig/editorconfig-core-go/pull/16 Even though, I might not be the correct way to handle the `unset` option.

Indeed. I should probably add more. Any pointers are appreciated.

### DIFF
--- a/struct.go
+++ b/struct.go
@@ -205,6 +205,20 @@ func setWithProperType(t reflect.Type, key *Key, field reflect.Value, delim stri
 		field.Set(reflect.ValueOf(timeVal))
 	case reflect.Slice:
 		return setSliceWithProperType(key, field, delim, allowShadow, isStrict)
+	case reflect.Ptr:
+		isBoolPtr := t.Elem().Kind() == reflect.Bool
+		if isBoolPtr {
+			boolVal, err := key.Bool()
+			if err != nil {
+				return wrapStrictError(err, isStrict)
+			}
+
+			v := reflect.New(t.Elem())
+			v.Elem().SetBool(boolVal)
+			field.Set(v)
+		} else {
+			return fmt.Errorf("unsupported type '%s'", t)
+		}
 	default:
 		return fmt.Errorf("unsupported type '%s'", t)
 	}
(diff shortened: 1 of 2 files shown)


struct.go
```diff
@@ -205,6 +205,20 @@ func setWithProperType(t reflect.Type, key *Key, field reflect.Value, delim stri
...
+			if err != nil {
+				return wrapStrictError(err, isStrict)
+			}
+
+			v := reflect.New(t.Elem())
+			v.Elem().SetBool(boolVal)
+			field.Set(v)
+		} else {
+			return fmt.Errorf("unsupported type '%s'", t)
+		}
```
unknwon: I think `reflect.Ptr` is basically no-type... it is OK we only support `*bool` right now. But for future support of other types, it would be better we use type switch here:

```go
switch t.Elem().Kind() {
	case reflect.Bool:
	default:
}
```

What do you think?

struct.go
```diff
@@ -424,6 +438,15 @@ func reflectWithProperType(t reflect.Type, key *Key, field reflect.Value, delim
...
 		return reflectSliceWithProperType(key, field, delim, allowShadow)
+	case reflect.Ptr:
+		isBoolPtr := t.Elem().Kind() == reflect.Bool
+		if isBoolPtr {
+			if !field.IsNil() {
+				key.SetValue(fmt.Sprint(field.Elem().Bool()))
+			}
+		} else {
+			return fmt.Errorf("unsupported type '%s'", t)
+		}
```
unknwon: See [the other comment](https://github.com/go-ini/ini/pull/200/files#r309976890).

struct_test.go
```diff
@@ -148,6 +151,7 @@ func Test_MapToStruct(t *testing.T) {
 			So(ts.Name, ShouldEqual, "Unknwon")
 			So(ts.Age, ShouldEqual, 21)
 			So(ts.Male, ShouldBeTrue)
+			So(ts.Optional, shouldEqualPtr, true)
```
unknwon: Would you mind explain why `ShouldEqual` won't work here?

struct.go
greut: Agreed.

struct_test.go
greut: ShouldEqual will compare the address value. I've changed it for `So(*ts.Optional, ShouldBeTrue)`

unknwon:
Thank you!

unknwon:
FYI, I've changed merge base to `v1` branch since this package now has Go modules enabled. I also tagged a new version for this merge: `v1.45.0`.

greut:
Cheers!

unknwon:
Just rolled back to disable Go modules :joy: too much troubles caused by enabling it.

PR #203, opened 2019-09-06
customization output delimiter

Lasiar:
Added customization output delimiter


ini.go
```diff
@@ -109,6 +109,8 @@ type LoadOptions struct {
 	UnparseableSections []string
 	// KeyValueDelimiters is the sequence of delimiters that are used to separate key and value. By default, it is "=:".
 	KeyValueDelimiters string
+	// KeyValueDelimiters is the delimiter that are used to separate key and value output. By default, it is "=".
+	KeyValueDelimitersOutput string
```
unknwon:
```suggestion
	KeyValueDelimiterOnWrite string
```

PR #206, opened 2019-09-15
Add support for []bool in structs

jeffa5:

This should add support for  the `[]bool` type in structs.

Yes.

### DIFF
--- a/key.go
+++ b/key.go
@@ -554,6 +554,12 @@ func (k *Key) Uint64s(delim string) []uint64 {
 	return vals
 }
 
+// Bools returns list of bool divided by given delimiter. Any invalid input will be treated as zero value.
+func (k *Key) Bools(delim string) []bool {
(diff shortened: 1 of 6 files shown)


jeffa5:
Fixed the failing tests.

key.go
```diff
@@ -398,6 +398,18 @@ func (k *Key) InUint64(defaultVal uint64, candidates []uint64) uint64 {
 	return defaultVal
 }
 
+// InBool always returns value without error,
+// it returns default value if error occurs or doesn't fit into candidates.
+func (k *Key) InBool(defaultVal bool, candidates []bool) bool {
```
unknwon: Thanks for adding these!

But this method seems particularly not useful in practice because boolean only have two values: true and false. What do you think?

jeffa5: Yeah, I agree, removing.

jeffa5:
Thanks!

unknwon:
Thank you!

PR #210, opened 2019-10-04
Large python style multiline value fix

r3db34n:
This change
- fixes issue #205
- supports large values (>4kB)
- allows to configure the maximum supported value size
- allows to turn on debug output to figure out why a multiline value does not parse correctly

There is a new test file with test cases as well as a test data file to cover empty lines in Python-style multiline values.


r3db34n:
Builds succeeds on all go versions except 1.9.x and 1.10.x because of test harness dependency `goconvey` (used by `go-ini/ini`): 
```
$ go get github.com/smartystreets/goconvey
# golang.org/x/tools/go/gcexportdata
../../../golang.org/x/tools/go/gcexportdata/gcexportdata.go:103:12: undefined: gcimporter.IExportData
The command "go get github.com/smartystreets/goconvey" exited with 2.
```

ini.go
```diff
@@ -29,7 +29,7 @@ const (
 
 	// Maximum allowed depth when recursively substituing variable names.
 	depthValues = 99
-	version     = "1.48.0"
+	version     = "1.48.0-1"
```
unknwon: Version will be changed after merge, no need to do it in advance.

ini_multiline_value_fix_test.go
```diff
@@ -0,0 +1,60 @@
+package ini_test
+
+import (
+	"gopkg.in/ini.v1"
+	. "github.com/smartystreets/goconvey/convey"
+	"path/filepath"
+	"testing"
```
unknwon: Please refer to other source files to spilt these into two groups.

ini_multiline_value_fix_test.go
```diff
@@ -0,0 +1,60 @@
+package ini_test
```
unknwon: I think `ini_python_multiline_test.go` is a better name.

ini_multiline_value_fix_test.go
```diff
@@ -0,0 +1,60 @@
...
+	Value1 := "some text here\n"+
+		"some more text here\n"+
+		"\n"+
+		"there is an empty line above and below\n"
+
+	Value2 := "there is an empty line above\n"+
+		"that is not indented so it should not be part\n"+
+		"of the value"
+
+	Convey("Parse Python-style multiline values", t, func() {
```
unknwon: Comments inside this func feel too verbose, please refer to tests in other source files to clean up a bit.

ini_multiline_value_fix_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+		// enable Python-style multiline values
+		opts := ini.LoadOptions{
+			AllowPythonMultilineValues: true,
+		}
+
+		// load test data
+		path := filepath.Join("testdata", "multiline.ini")
+		data, e := ini.LoadSources(opts, path)
```
unknwon:
```suggestion
		f, err := ini.LoadSources(opts, path)
```

ini_multiline_value_fix_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	Convey("Parse Python-style multiline values", t, func() {
+
+		// enable Python-style multiline values
+		opts := ini.LoadOptions{
+			AllowPythonMultilineValues: true,
+		}
+
+		// load test data
+		path := filepath.Join("testdata", "multiline.ini")
```
unknwon: Static test data can be defined use a `const`, unless there is a good reason that we must interact with file system.

ini_multiline_value_fix_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func TestMultiline(t *testing.T) {
+	Value1 := "some text here\n"+
+		"some more text here\n"+
+		"\n"+
+		"there is an empty line above and below\n"
+
+	Value2 := "there is an empty line above\n"+
+		"that is not indented so it should not be part\n"+
+		"of the value"
```
unknwon: These two values only be used once, at equality check, I think directly put string literals in `So(...` is fine.

parser.go
```diff
@@ -47,8 +50,13 @@ type parser struct {
...
 
 func newParser(r io.Reader, opts parserOptions) *parser {
+	size := opts.ReaderBufferSize
+	if size < minReaderBufferSize {
+		size = minReaderBufferSize
+	}
+
 	return &parser{
-		buf:     bufio.NewReader(r),
+		buf:     bufio.NewReaderSize(r, size),
```
unknwon: What is the rationale to use `bufio.NewReaderSize` rather than `bufio.NewReader`? What is not working with `bufio.NewReader`?

unknwon:
I don't know what's going with 1.9 and 1.10, we can ignore it for now.

r3db34n:
cool

r3db34n:
See my comments below.

unknwon:
Hi, thanks for the follow up. Can you please comment directly on the GitHub UI?

r3db34n:
Ok, will do tomorrow. It is getting very late here.

unknwon:
No problem, it's no hurry, good night!

ini.go
r3db34n: I will roll back this change as requested.

ini_multiline_value_fix_test.go
r3db34n: I will adopt this change as requested.

ini_multiline_value_fix_test.go
r3db34n: I will adopt this change as requested.

parser.go
r3db34n: I needed to read large values. However, bufio.NewReader() limits the internal buffer size to 4kB. 

I needed this size to be configurable, so I added a config option for it and passed it to NewReaderSize(). Hence the ‘large’ in the title of the pull request ;)

ini_multiline_value_fix_test.go
r3db34n: As a user, I find it easier to see actual examples of ini files that are parsed correctly in the `testdata` folder rather than to have to read test cases to find them. I would prefer to keep the test data in the separate ini file.

ini_multiline_value_fix_test.go
r3db34n: Which comments?

ini_multiline_value_fix_test.go
r3db34n: Sure, no problem

ini_multiline_value_fix_test.go
r3db34n: OK, done

ini_multiline_value_fix_test.go
unknwon: Hmm, I would say all the comments in the test convey.

ini_multiline_value_fix_test.go
unknwon: AFA I see, the content of "multiline.ini" is too simple and straightforward, not worth keeping it in a file unless the purpose of test is to test interactions with file system (i.e. read/write a file).

unknwon: For good examples to users, documentation is the better place.

ini_python_multiline_test.go
```diff
@@ -0,0 +1,50 @@
...
+	Convey("Parse Python-style multiline values", t, func() {
+
+		// enable Python-style multiline values
+		opts := ini.LoadOptions{
+			AllowPythonMultilineValues: true,
+		}
+
+		// load test data
+		path := filepath.Join("testdata", "multiline.ini")
+		f, err := ini.LoadSources(opts, path)
```
unknwon: No need to define `opts` since it is only used once, and to match style in every other places of tests.

```go
f, err := ini.LoadSources(ini.LoadOptions{
	AllowPythonMultilineValues: true,
}, path)
```

ini_python_multiline_test.go
```diff
@@ -0,0 +1,50 @@
...
+		defaultSection := f.Section("")
+		So(f.Section(""), ShouldNotBeNil)
+
+		// Default section should map to test data struct
+		var testData testData
+		err = defaultSection.MapTo(&testData)
+		So(err, ShouldBeNil)
+
+		// Parsed values should match expected values
+		So(testData.Value1, ShouldEqual, "some text here\nsome more text here\n\nthere is an empty line above and below\n")
```
unknwon: If you think using `\n` is not very obvious, you could use back quotes:

```go
So(testData.Value1, ShouldEqual, `some text here
some more text here

there is an empty line above and below
`
```

parser.go
```diff
@@ -25,7 +25,8 @@ import (
 	"unicode"
 )
 
-var pythonMultiline = regexp.MustCompile("^(\\s+)([^\n]+)")
+const minReaderBufferSize = 64*1024
```
unknwon: I would suggest use 4096 (4KB, which is the size `bufio.NewReader` uses), which reminds default unchanged. 

User should set `ReaderBufferSize` to a larger number when he/she needs to (in your case, set it to 64KB when your program loads the INI file).

parser.go
```diff
@@ -25,7 +25,8 @@ import (
...
+var pythonMultiline = regexp.MustCompile(`^([\t\f ]+)(.*)`)
```
unknwon: Would you mind explain what has changed in this regex? (I'm a rookie about regex)

parser.go
```diff
@@ -285,33 +293,76 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+
 		peekMatches := pythonMultiline.FindStringSubmatch(string(peekData))
+		if p.options.Debug {
+			fmt.Println("readPythonMultilines: matched ", len(peekMatches), " parts:")
+			for n, v := range peekMatches {
+				fmt.Println("   ", n, ": '", v, "'")
+			}
+		}
+
+		// return if not a python-ini multi-line value.
```
unknwon:
```suggestion
		// Return if not a Python multiline value.
```

parser.go
```diff
@@ -285,33 +293,76 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+				fmt.Println("readPythonMultilines: didn't match enough parts, meaning the value has ended")
+				fmt.Println("readPythonMultilines: value is: '"+line+"'")
+			}
 			return line, nil
 		}
 
-		// NOTE: Return if not a python-ini multi-line value.
-		currentIdentSize := len(peekMatches[1])
-		if currentIdentSize <= 0 {
+		// determine indent size and line prefix
```
unknwon:
```suggestion
		// Determine indent size and line prefix.
```

parser.go
```diff
@@ -285,33 +293,76 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+		currentIndentSize := len(peekMatches[1])
+		if indentSize < 1 {
+			indentSize = currentIndentSize
+			if p.options.Debug {
+				fmt.Println("readPythonMultilines: new indent size is ", indentSize)
+			}
+		}
+
+		// make sure each line is indented at least as far as first line
```
unknwon:
```suggestion
		// Make sure each line is indented at least as far as first line.
```

parser.go
```diff
@@ -285,33 +293,76 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+				fmt.Println("readPythonMultilines: current indent size is ", currentIndentSize)
+				fmt.Println("readPythonMultilines: current line is not indented enough to be part of this value")
+				fmt.Println("readPythonMultilines: value is: '"+line+"'")
+			}
 			return line, nil
 		}
 
-		// NOTE: Just advance the parser reader (buffer) in-sync with the peek buffer.
-		_, err := p.readUntil('\n')
+		// advance the parser reader (buffer) in-sync with the peek buffer.
```
unknwon:
```suggestion
		// Advance the parser reader (buffer) in-sync with the peek buffer.
```

parser.go
```diff
@@ -285,33 +293,76 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+		_, err := p.buf.Discard(len(peekData))
 		if err != nil {
+			if p.options.Debug {
+				fmt.Println("readPythonMultilines: failed to skip to the end, returning error")
+			}
 			return "", err
 		}
 
-		line += fmt.Sprintf("\n%s", peekMatches[2])
+		// handle indented empty line
```
unknwon:
```suggestion
		// Handle indented empty line.
```

parser.go
```diff
@@ -348,8 +401,8 @@ func (f *File) parse(reader io.Reader) (err error) {
 	// the size of the parser buffer is found.
 	// TODO(unknwon): When Golang 1.10 is the lowest version supported, replace with `parserBufferSize := p.buf.Size()`.
 	parserBufferSize := 0
-	// NOTE: Peek 1kb at a time.
-	currentPeekSize := 1024
+	// NOTE: Peek 40kb at a time.
```
unknwon:
```suggestion
	// NOTE: Peek 4KB at a time.
```

ini_multiline_value_fix_test.go
r3db34n: I added another test case to cover large values. I think this should be in a separate test file.

ini_python_multiline_test.go
r3db34n: OK, adopted

parser.go
r3db34n: OK, adopted

parser.go
r3db34n: OK, adopted

parser.go
r3db34n: OK, adopted

parser.go
r3db34n: OK, adopted

parser.go
r3db34n: OK, adopted

parser.go
r3db34n: OK, adopted

parser.go
r3db34n: OK, adopted

ini_python_multiline_test.go
r3db34n: I wanted the tests to be in a single line. Now that I added a new test case with very large text, I used this method above for that. I am not sure if you are asking me to change the other tests to follow the same pattern. To make it simple, I will go ahead and change them as well just in case to close out this pull request

parser.go
r3db34n: \s is equivalent to [\r\n\t\f\v ], e.g. not includes newlines and carriage returns. that is a problem when trying to parse a multiline value one line at a time. It matched empty lines without indents which made them part of the value when they shouldn't. Multiline values are terminated when a line is encountered that has a lesser whitespace indent than the first line after the line with the key. By using \s it was not possible to catch the termination condition correctly.

unknwon: Thanks for the explanation!

unknwon:
Looks like CI is passing now, maybe was the flakiness of Travis CI.

r3db34n:
Is there anything else you need me to do to accept the pull request?

ini.go
```diff
@@ -111,6 +111,10 @@ type LoadOptions struct {
 	KeyValueDelimiters string
 	// PreserveSurroundedQuote indicates whether to preserve surrounded quote (single and double quotes).
 	PreserveSurroundedQuote bool
+	// turn on debug output (currently only useful to debug parsing Python-style multiline values)
```
unknwon:
```suggestion
	// Debug indicates whether to collect debug information (currently only useful to debug parsing Python-style multiline values).
```

ini.go
```diff
@@ -111,6 +111,10 @@ type LoadOptions struct {
...
+	Debug bool
+	// increase reader buffer size
```
unknwon:
```suggestion
	// ReaderBufferSize is the buffer size of the reader in bytes.
```

parser.go
```diff
@@ -346,10 +399,10 @@ func (f *File) parse(reader io.Reader) (err error) {
 
 	// NOTE: Iterate and increase `currentPeekSize` until
 	// the size of the parser buffer is found.
-	// TODO(unknwon): When Golang 1.10 is the lowest version supported, replace with `parserBufferSize := p.buf.Size()`.
+	// TODO(unknown): When Golang 1.10 is the lowest version supported, replace with `parserBufferSize := p.buf.Size()`.
```
unknwon: This change is not need, that is my correct GitHub handle :)

parser.go
```diff
@@ -285,33 +293,76 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
 	parserBufferPeekResult, _ := p.buf.Peek(bufferSize)
 	peekBuffer := bytes.NewBuffer(parserBufferPeekResult)
 
+	indentSize := 0
 	for {
 		peekData, peekErr := peekBuffer.ReadBytes('\n')
 		if peekErr != nil {
 			if peekErr == io.EOF {
+				if p.options.Debug {
+					fmt.Println("readPythonMultilines: failed to peek")
```
unknwon: Although I understand you would like to have debug info to help debug, but I don't think print to stdout is suitable for this package, i.e. this is not a Go program but a utility package. 

With that being said, I think debug printing should not be committed for this particular package.

unknwon: Alternatively, we can change `Debug bool` to `DebugWriter io.Writer`, and if it's not nil, write debug info to this writer.

Then we just need to think about what to write.

unknwon:
This PR is ready to merge without the "debug thing". Thanks again for the work, but it's your call :)

parser.go
r3db34n: Sorry, I didn't realize. will roll back

parser.go
r3db34n: In my projects, I use loggers like "github.com/sirupsen/logrus". This allows anyone using your project to configure logging any way they want. not sure if you want to do that here. If so, I can modify the code to use that package

ini.go
r3db34n: will do

ini.go
r3db34n: will adopt

r3db34n:
> This PR is ready to merge without the "debug thing". Thanks again for the work, but it's your call :)

I will make those changes you requested. Let me know how you want to proceed with the logging

parser.go
unknwon: Honestly, I don't want to introduce debugging feature and/or another dependency because I'm not convinced this will have enough benefit.

r3db34n: OK, I made it a debug function. Will commit changes soon

r3db34n:
OK, all changes have been made as requested. Should be good to go now

ini.go
```diff
@@ -111,8 +111,15 @@ type LoadOptions struct {
 	KeyValueDelimiters string
 	// PreserveSurroundedQuote indicates whether to preserve surrounded quote (single and double quotes).
 	PreserveSurroundedQuote bool
+	// Debug indicates whether to collect debug information (currently only useful to debug parsing Python-style multiline values).
+	Debug DebugFunc
+	// ReaderBufferSize is the buffer size of the reader in bytes.
+	ReaderBufferSize int
 }
 
+// DebugFunc is the type of function called to log parse events
```
unknwon:
```suggestion
// DebugFunc is the type of function called to log parse events.
```

ini.go
```diff
@@ -111,8 +111,15 @@ type LoadOptions struct {
...
+type DebugFunc func(args ...interface{})
```
unknwon: How does a user know which arg corresponding to which?

unknwon: I still think `DebugWriter io.Writer` is the cleanest solution here.

r3db34n: args should be passed to something like fmt.Println(). A typical implementation would look like this:
```
func Debug(args ...interface{}) {
   fmt.Println(args...)
}
```
This allows to easily plug in any logger or writer.

ini.go
r3db34n: adopted

ini_python_multiline_test.go
```diff
@@ -0,0 +1,65 @@
...
+	Convey("Parse Python-style multiline values", t, func() {
+		path := filepath.Join("testdata", "multiline.ini")
+		f, err := ini.LoadSources(ini.LoadOptions{
+			AllowPythonMultilineValues: true,
+			ReaderBufferSize: 64*1024,
+			/*
+			Debug: func(m string) {
+				fmt.Println(m)
+			},
+			 */
```
unknwon: Why is this comment necessary?

parser.go
```diff
@@ -35,6 +36,8 @@ type parserOptions struct {
 	UnescapeValueDoubleQuotes   bool
 	UnescapeValueCommentSymbols bool
 	PreserveSurroundedQuote     bool
+	Debug 						DebugFunc
```
unknwon: I think this field should also be called `DebugFunc`, because it is a function now.

ini.go
```diff
@@ -111,8 +111,15 @@ type LoadOptions struct {
 	KeyValueDelimiters string
 	// PreserveSurroundedQuote indicates whether to preserve surrounded quote (single and double quotes).
 	PreserveSurroundedQuote bool
+	// Debug indicates whether to collect debug information (currently only useful to debug parsing Python-style multiline values).
+	Debug DebugFunc
```
unknwon: I think this field should also be called `DebugFunc`, because it is a function now.

ini.go
```diff
@@ -111,8 +111,15 @@ type LoadOptions struct {
 	KeyValueDelimiters string
 	// PreserveSurroundedQuote indicates whether to preserve surrounded quote (single and double quotes).
 	PreserveSurroundedQuote bool
+	// Debug indicates whether to collect debug information (currently only useful to debug parsing Python-style multiline values).
```
unknwon:
```suggestion
	// DebugFunc is called to collect debug information (currently only useful to debug parsing Python-style multiline values).
```

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
 	peekBuffer := bytes.NewBuffer(parserBufferPeekResult)
 
+	indentSize := 0
 	for {
 		peekData, peekErr := peekBuffer.ReadBytes('\n')
 		if peekErr != nil {
 			if peekErr == io.EOF {
+				p.debug("readPythonMultilines: failed to peek")
+				p.debug("readPythonMultilines: peekData is: '"+string(peekData)+"'")
+				p.debug("readPythonMultilines: value is: '"+line+"'")
```
unknwon:
```suggestion
				p.debug(fmt.Sprintf("parser.readPythonMultilines: io.EOF, peekData: %q, line: %q", string(peekData), line))
```

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
 				return line, nil
 			}
+			p.debug("readPythonMultilines: failed to peek, returning error")
```
unknwon: Why this is necessary since `peekErr` is returned?

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
 			return "", peekErr
 		}
 
+		p.debug("readPythonMultilines: parsing '"+string(peekData)+"'")
```
unknwon:
```suggestion
		p.debug(fmt.Sprintf("parser.readPythonMultilines: parsing %q", string(peekData)))
```

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+
 		peekMatches := pythonMultiline.FindStringSubmatch(string(peekData))
+		p.debug("readPythonMultilines: matched ", len(peekMatches), " parts:")
+		for n, v := range peekMatches {
+			p.debug("   ", n, ": '", v, "'")
+		}
```
unknwon: One single debug line should be self-contained, it can be very long and I don't think that will be a problem.

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+
+		// Return if not a Python multiline value.
 		if len(peekMatches) != 3 {
+			p.debug("readPythonMultilines: didn't match enough parts, meaning the value has ended")
+			p.debug("readPythonMultilines: value is: '"+line+"'")
```
unknwon: One single debug line should be self-contained, it can be very long and I don't think that will be a problem.

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
 		}
 
-		// NOTE: Return if not a python-ini multi-line value.
-		currentIdentSize := len(peekMatches[1])
-		if currentIdentSize <= 0 {
+		// Determine indent size and line prefix.
+		currentIndentSize := len(peekMatches[1])
+		if indentSize < 1 {
+			indentSize = currentIndentSize
+			p.debug("readPythonMultilines: new indent size is ", indentSize)
```
unknwon:
```suggestion
			p.debug("parser.readPythonMultilines: new indent size is", indentSize)
```

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
+		}
+
+		// Make sure each line is indented at least as far as first line.
+		if currentIndentSize < indentSize {
+			p.debug("readPythonMultilines: expected indent size is ", indentSize)
+			p.debug("readPythonMultilines: current indent size is ", currentIndentSize)
+			p.debug("readPythonMultilines: current line is not indented enough to be part of this value")
+			p.debug("readPythonMultilines: value is: '"+line+"'")
```
unknwon: One single debug line should be self-contained, it can be very long and I don't think that will be a problem.

parser.go
```diff
@@ -285,33 +299,60 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
 			return line, nil
 		}
 
-		// NOTE: Just advance the parser reader (buffer) in-sync with the peek buffer.
-		_, err := p.readUntil('\n')
+		// Advance the parser reader (buffer) in-sync with the peek buffer.
+		_, err := p.buf.Discard(len(peekData))
 		if err != nil {
+			p.debug("readPythonMultilines: failed to skip to the end, returning error")
```
unknwon: Why this is necessary since `err` is returned?

ini_python_multiline_test.go
r3db34n: This is what I used to debug the code change. It is also an example for someone who wants to know how to use it. It is in the test case, not the core module, so I don't think it is an issue.

parser.go
r3db34n: The debug line in the for loop can't be expressed any simpler since at this point in the function it is unknown how many elements are in peekMatches.

parser.go
r3db34n: err does not contain that information. this is useful to figure our why and where parsing failed

parser.go
r3db34n: makes sense

ini.go
r3db34n: makes sense

ini.go
r3db34n: makes sense

parser.go
r3db34n: I don't think this is necessary, but whatever, I'll make the change to close this out

parser.go
r3db34n: I don't think this is necessary, but whatever, I'll make the change to close this out

parser.go
r3db34n: makes sense

parser.go
r3db34n: I don't think this is necessary, but whatever, I'll make the change to close this out

r3db34n:
I will make those changes as agreed, but this will be my final round of modifications. If you want further changes, I would suggest you merge the code and make the changes in your repo.

parser.go
unknwon: Ping on this comment, possible missed.

parser.go
unknwon: > since at this point in the function it is unknown how many elements are in peekMatches.

What do you mean by "unknown how many elements in peekMatches"? `len(peekMatches)`?

unknwon:
I feel GitHub review is broken... so I'll just list links of two comments:

- https://github.com/go-ini/ini/pull/210/files#r331858607
- https://github.com/go-ini/ini/pull/210/files#r341452664

parser.go
r3db34n: I was referring to your comment that you want this to be in a single line. I responded saying that it can't be since I have to iterate over the number of matches, hence the for loop instead of a single line. If the number of matches was fixed, I could just write a single statement printing all at once.

parser.go
r3db34n: This is to provide an easy way to trace what exactly is failing and why. The debug() function should provide additional information beyond the thrown exception. When implementing debug logging, I usually have a debug log statement at every escape.

unknwon: OK, make sense, then I think probably could do this:
```suggestion
			p.debug("readPythonMultilines: failed to peek with error: %v", err)
```

r3db34n: ok

parser.go
r3db34n: done

unknwon:
Thank you! Merging.

PR #213, opened 2019-11-06
Migrate to Go modules

jpreese:
Importing this package requires a dependency on `gopkg.ini` and currently lacks a `go.mod` file to highlight its dependency on `goconvey`

### DIFF
--- a/bench_test.go
+++ b/bench_test.go
@@ -17,7 +17,7 @@ package ini_test
 import (
 	"testing"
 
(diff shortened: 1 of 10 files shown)


unknwon:
Hi, I've tried, but given this package is heavily used by two import paths `gopkg.in/ini.v1` and `github.com/go-ini-ini`, Go Modules can't handle that.

Will enable Go Modules in v2.

PR #214, opened 2019-11-08
parser: support unmashalling to all basic type pointers

jstewmon:

Unmashalling to a pointer field facilitates distinguishing between a key being undefined and being defined with the type's zero value. This is useful for cases where an application defines a default value, which is not the type's zero value and the config should be able to set the field value to the type's zero value.

This feature was already supported for `*bool` - I have extended the capability to all of the supported basic types.

Yes, I have extended the existing test to cover all of the new types.

### DIFF
--- a/ini_python_multiline_test.go
+++ b/ini_python_multiline_test.go
@@ -4,8 +4,8 @@ import (
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


unknwon:
Thank you!

Issue #218, opened 2019-12-20
Incompatibility error with older versions


YangKian:
### Describe the bug

When I update go-ini to the latest version, I found some error, here is my case:
```go
//my.ini file
[server]
ReadTimeout = 60

//main.go file
import (
	"github.com/go-ini/ini"
	"time"
)

type Server struct {
    ReadTimeout time.Duration
}

var (
    	ServerSetting = &Server{}
    	cfg *ini.File
)

func main() {
	var err error
	if cfg, err = ini.Load("conf/app.ini"); err != nil {
		log.Fatalf("Fail to parse 'conf/app.ini': %v\n", err)
	}

	if err := cfg.Section("server").MapTo(ServerSetting); err != nil {
		log.Fatalf("Cfg.MapTo %s err: %v\n", "server", err)
	}

	ServerSetting.ReadTimeout = ServerSetting.ReadTimeout * time.Second
	fmt.Println(ServerSetting.ReadTimeout)
}
```

For old versions, the result of `fmt.Println(ServerSetting.ReadTimeout)` was `60s`, but for the latest version I get a result of `0s`.

### More informations about the bug

I did some trace and found the err happend here :

![err1](https://user-images.githubusercontent.com/45479280/71264286-bc496700-237e-11ea-9578-93b5a3732e60.png)

Then I search the merge history and found :

![history](https://user-images.githubusercontent.com/45479280/71264306-c53a3880-237e-11ea-9b22-360737272d7c.png)

In the old version, my case will get an err in ` part 1`  too, but it still had another chance to set the correct value in `part 2` and `part 3`，but the new version will return an err directly in `part 4`, that's the reason why wrong result returned.

YangKian:
I was able to fix it by adding a nest `if case` in `part 4`  :
![mytry](https://user-images.githubusercontent.com/45479280/71265419-2531de80-2381-11ea-8ccd-6d9b2dee3ed3.png)

PR #219, opened 2019-12-20
struct: fix incompatibility error with older versions

YangKian:
This fixes issue #218 

Yes，I add a new test case to cover the situation issue #218 mentioned

### DIFF
--- a/struct.go
+++ b/struct.go
@@ -183,6 +183,10 @@ func setWithProperType(t reflect.Type, key *Key, field reflect.Value, delim stri
 		if vt.Name() == "Duration" {
 			durationVal, err := key.Duration()
 			if err != nil {
+				if intVal, err := key.Int64(); err == nil {
+					field.SetInt(intVal)
+					return nil
+				}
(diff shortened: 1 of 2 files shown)


struct.go
```diff
@@ -183,6 +183,10 @@ func setWithProperType(t reflect.Type, key *Key, field reflect.Value, delim stri
 		if vt.Name() == "Duration" {
 			durationVal, err := key.Duration()
 			if err != nil {
+				if intVal, err := key.Int64(); err == nil {
+					field.SetInt(intVal)
+					break
```
unknwon: I think this is more appropriate:
```suggestion
					return nil
```

YangKian: Yes, its a better way to use `return nil`, I have fix it, thanks a lot :) So sorry for submit too many commits, I'm green hands for git, tried to merge and modify the commits, but failed.

unknwon: @YangKian No worries! GitHub has squash merge so it doesn't matter :)

unknwon:
Thank you!

Issue #190, opened 2019-05-06
non unique sections


aligator:
### Please give general description of the problem
Wireguard has ini files like this:
[Interface]
Address = 192.168.2.1
PrivateKey = <server's privatekey>
ListenPort = 51820

[Peer]
PublicKey = <client's publickey>
AllowedIPs = 192.168.2.2/32

[Peer]
PublicKey = <client2's publickey>
AllowedIPs = 192.168.2.3/32

And there can be unlimited [Peer] sections.
Is it currently possible to parse / write such an ini-file?

### Do you have any suggestion to fix the problem?
I haven't found anything about this so I already tried it myself: https://github.com/aligator/ini/tree/nonUniqueSections
This already works for ini.LoadSources and f.WriteTo but I haven't tested it very much.
Edit: now for ini.MapTo also (but not f.Section("Peer").MapTo)

And I don't know where else in the source I have to add support for this...
If you have any hints for me it would be great :-)

Unknwon:
Hi, glad you got it working! You could send a PR from your fork to this repository, I will review but I can't promise it would be merged in advance.

aligator:
It's not finished yet, and currently I have some other projects I work on. Don't know when I work on this again. (It's on my todo-list :-) Should I create a WIP PR so that it doesn't get lost if someone else needs this functionality or wants to finish the work?

Unknwon:
@aligator no pressures!

unknwon:
Hey, what is the status of this?

aligator:
Oh, sorry, I am still working on other projects and as always I have not enough time... I haven't worked on this yet.

unknwon:
Thanks for the info! No worries, just checking.

Place1:
I've run into the same limitation, trying to parse the same wireguard config as well haha. I'd be keen to see support for this.

vx3r:
anything new about this issue ?

PR #220, opened 2020-01-23
file: support non-unique sections

aligator:
First of all: this is not finished yet.
It fixes https://github.com/go-ini/ini/issues/190

As there seem to be some more people now who need this, I would like to continue the work now.

I create this PR because I need some feedback if it is ok how I implement it.

Currently I have implemented ini.LoadSources(), f.WriteTo(), f.DeleteSpecificSection(), ini.MapTo(), f.Section("Peer").MapTo()

Wireguard has ini files like this:
```
[Interface]
Address = 192.168.2.1
PrivateKey = <server's privatekey>
ListenPort = 51820

[Peer]
PublicKey = <client's publickey>
AllowedIPs = 192.168.2.2/32

[Peer]
PublicKey = <client2's publickey>
AllowedIPs = 192.168.2.3/32
```

And there can be unlimited [Peer] sections.

These non unique sections are currently not supported.

Yes, for all cases I have currently implemented. There are still cases missing to implement and to write tests for.


Neo-Desktop:
@aligator Is this PR finished?

aligator:
I am not sure if I found all constellations where unique sections could be possible.
I think @unknwon can review it for now. If there is still something missing, I will have a look.

unknwon:
Thanks for the PR! Could you have a paragraph to explain the design ideas before dive into the code?

aligator:
Yes of course:

First of all I tried to not break any current functionality.

In file.go I replaced the sections map[string]*Section by a sections map[string][]*Section

So that each section can be added multiple times.

So if we look at the following test-ini:
```
[Interface]
Address = 192.168.2.1
PrivateKey = <server's privatekey>
ListenPort = 51820

[Peer]
PublicKey = <client's publickey>
AllowedIPs = 192.168.2.2/32

[Peer]
PublicKey = <client2's publickey>
AllowedIPs = 192.168.2.3/32
```

loaded using
ini.LoadSources(ini.LoadOptions{AllowNonUniqueSections: true,}, ...

sections contains a list of all occuring section names and each of them can have several *section:
```

{
	"DEFAULT": [ sec1 ],
	"Interface": [ sec2 ],
	"Peer": [ sec3, sec4 ]
}
```

As you can see Peer now just has two sections.

The sectionList now can contain the same name two times.
```

sectionList = {[]string} len:4, cap:10
 0 = {string} "DEFAULT"
 1 = {string} "Interface"
 2 = {string} "Peer"
 3 = {string} "Peer"
```
 
To keep track of the positions where each instance is stored in sections, I added the sectionNumList.
 
```
sectionNumList = {[]int} len:4, cap:10
 0 = {int} 0
 1 = {int} 0
 2 = {int} 0
 3 = {int} 1
```
 
so this two lists together tell exactly where to find the instances:
```
DEFAULT is found in 
sections["DEFAULT"][0]

Interface is found in 
sections["Interface"][0]

the first Peer is found in 
sections["Peer"][0]

the second Peer is found in 
sections["Peer"][1]

```
Of course these lists have to be updated each time, a new section is added or one is removed.

Note that for backwards compatibility f.GetSection only returns the first section with that name.
If you would like to get all of that name call f.GetSections.

I also added a GetSectionCount and SpecificSection. The last one can be used for example to obtain only the second section named "Peer".

If you remove a section using DeleteSection all sections are deleted with the given name.
Thats why I also added DeleteSpecificSection.

For loading the non unique sections to a struct I added the allowNonUnique annotation:
```
type testNonUniqueSectionsStruct struct {
	Interface testInterface
	Peer      []testPeer `ini:",,,allowNonUnique"`
}
```

I think the best starting point for you would be the tests I added.

unknwon:
Thanks @aligator, very helpful!

ini.go
```diff
@@ -134,6 +136,12 @@ func LoadSources(opts LoadOptions, source interface{}, others ...interface{}) (_
 	return f, nil
 }
 
+func LoadCustomEmpty(opts LoadOptions) *File {
```
unknwon: I think it's cleaner to just change the following function to accept an optional `LoadOptions` via variable-length arguments, i.e. 

```go
func Empty(opts ... LoadOptions) *File {
	var opt LoadOptions
	if len(opts) > 0 {
		opt = opts[0]
	}

	...
}
```

https://github.com/go-ini/ini/blob/dd614b817be88f32a238e94e84f11da890471a7c/file.go#L61

file.go
```diff
@@ -36,8 +36,10 @@ type File struct {
 
 	// To keep data in order.
 	sectionList []string
+	// To keep track of the number of a section, if non unique section-names are allowed
+	sectionNumList []int
```
unknwon:
```suggestion
	// To keep track of the index of a section with same name. 
	// This meta list is only used with non-unique section names are allowed.
	sectionIndexes []int
```

file.go
```diff
@@ -49,11 +51,12 @@ func newFile(dataSources []dataSource, opts LoadOptions) *File {
...
-		BlockMode:   true,
-		dataSources: dataSources,
-		sections:    make(map[string]*Section),
-		sectionList: make([]string, 0, 10),
-		options:     opts,
+		BlockMode:      true,
+		dataSources:    dataSources,
+		sections:       make(map[string][]*Section),
+		sectionList:    make([]string, 0, 10),
+		sectionNumList: make([]int, 0, 10),
```
unknwon: Shouldn't we only initialize this when `opts.AllowNonUniqueSections` is `true`?

file.go
```diff
@@ -64,6 +67,12 @@ func Empty() *File {
 	return f
 }
 
+func (f *File) reset(name string) {
```
unknwon: What's the use case of this method?

file.go
```diff
@@ -110,6 +121,17 @@ func (f *File) NewSections(names ...string) (err error) {
 
 // GetSection returns section by given name.
 func (f *File) GetSection(name string) (*Section, error) {
+	sec, err := f.GetSections(name)
+
```
unknwon:
```suggestion
	secs, err := f.GetSections(name)
```

file.go
```diff
@@ -123,12 +145,24 @@ func (f *File) GetSection(name string) (*Section, error) {
 	}
 
 	sec := f.sections[name]
```
unknwon:
```suggestion
	secs := f.sections[name]
```

file.go
```diff
@@ -123,12 +145,24 @@ func (f *File) GetSection(name string) (*Section, error) {
...
-	if sec == nil {
+	if sec == nil || len(sec) == 0 {
```
unknwon: `len` handles `nil`:
```suggestion
	if len(sec) == 0 {
```

file.go
```diff
@@ -123,12 +145,24 @@ func (f *File) GetSection(name string) (*Section, error) {
...
 		return nil, fmt.Errorf("section '%s' does not exist", name)
 	}
+
 	return sec, nil
 }
 
+// GetSectionCount gives you the count of sections with the same name.
+func (f *File) GetSectionCount(name string) int {
```
unknwon: What's the use case of this method?

file.go
```diff
@@ -141,6 +175,19 @@ func (f *File) Section(name string) *Section {
 	return sec
 }
 
+// Section assumes named section exists and returns a zero-value when not.
+func (f *File) SpecificSection(name string, pos int) *Section {
```
unknwon:
```suggestion
func (f *File) GetSectionWithIndex(name string, index int) *Section {
```

file.go
```diff
@@ -168,7 +215,17 @@ func (f *File) SectionStrings() []string {
...
 func (f *File) DeleteSection(name string) {
+	count := f.GetSectionCount(name)
+
+	for i := 0; i < count; i++ {
+		f.DeleteSpecificSection(name, 0)
+	}
+}
+
+// DeleteSection deletes a section.
+func (f *File) DeleteSpecificSection(name string, num int) {
```
unknwon:
```suggestion
func (f *File) DeleteSectionWithIndex(name string, index int) {
```

struct.go
```diff
@@ -263,16 +263,19 @@ func setWithProperType(t reflect.Type, key *Key, field reflect.Value, delim stri
...
 	rawName = opts[0]
 	if len(opts) > 1 {
 		omitEmpty = opts[1] == "omitempty"
 	}
 	if len(opts) > 2 {
 		allowShadow = opts[2] == "allowshadow"
 	}
-	return rawName, omitEmpty, allowShadow
+	if len(opts) > 3 {
+		allowNonUnique = opts[3] == "allowNonUnique"
```
unknwon: I think tag are all in lower cases:
```suggestion
		allowNonUnique = opts[3] == "nonunique"
```

struct.go
```diff
@@ -326,8 +341,34 @@ func (s *Section) mapTo(val reflect.Value, isStrict bool) error {
 	return nil
 }
 
-// MapTo maps section to given struct.
-func (s *Section) MapTo(v interface{}) error {
+// maps all sections with the same name and returns the new value.
+func (s *Section) mapAllTo(secName string, val reflect.Value, isStrict bool) (reflect.Value, error) {
```
unknwon: Please use defensive programming here, return early to save some indentations.

ini.go
aligator: Yes, good idea

file.go
aligator: I have to think about it and test it. I am not sure if it would break something because the code is now using these lists always. If the option is false there are just some errors thrown if the user tries to load non unique sections, if I remember correctly.

file.go
aligator: Don't know, I think I used it in an early version... I'll remove it.

file.go
aligator: It is used in DeleteSection and I thought that it may be a feature, someone else could also need. That's why I made it public. But as it's only a len on f.GetSections it would be also ok to remove it and put the code directly into DeleteSection, if you don't like it.

unknwon: Yeah, just embed in the `DeleteSection` SGTM.

struct.go
aligator: I am not sure what exactly you mean. Where can I return more early?

file.go
aligator: i added some checks and only initialize the index list if opts.AllowNonUniqueSections is true

struct.go
```diff
@@ -326,8 +341,34 @@ func (s *Section) mapTo(val reflect.Value, isStrict bool) error {
...
+				// add element to field
+				val = reflect.Append(val, newElement.Elem())
+			}
+			return val, nil
+		} else {
+			return reflect.Value{}, err
+		}
+	} else {
+		return reflect.Value{}, fmt.Errorf("error mapping section(%s): Val has to be of type Slice", secName)
+	}
```
unknwon:
```suggestion
	if typ.Kind() != reflect.Slice {
		return reflect.Value{}, fmt.Errorf("error mapping section(%s): Val has to be of type Slice", secName)
	}
	
	secs, err := s.f.GetSections(secName)
	if err != nil {
		return reflect.Value{}, err
	} 
	
	for _, sec := range secs {
		sliceType := typ.Elem()

		// create new element
		newElement := reflect.New(sliceType)
		if err = sec.mapTo(newElement, isStrict); err != nil {
			return reflect.Value{}, fmt.Errorf("error mapping section(%s): %v", secName, err)
		}

		// add element to field
		val = reflect.Append(val, newElement.Elem())
	}
	return val, nil
```

aligator: Ah, now I I understand what you mean. Thank you for the suggestion.

file.go
```diff
@@ -113,6 +133,17 @@ func (f *File) NewSections(names ...string) (err error) {
 
 // GetSection returns section by given name.
 func (f *File) GetSection(name string) (*Section, error) {
+	secs, err := f.GetSections(name)
+
```
unknwon:
```suggestion
```

file.go
```diff
@@ -126,9 +157,10 @@ func (f *File) GetSection(name string) (*Section, error) {
 	}
 
 	sec := f.sections[name]
-	if sec == nil {
+	if sec == nil || len(sec) == 0 {
 		return nil, fmt.Errorf("section '%s' does not exist", name)
```
unknwon: Would be nice have this fixed altogether in this PR:
```suggestion
		return nil, fmt.Errorf("section %q does not exist", name)
```

file.go
```diff
@@ -144,6 +176,19 @@ func (f *File) Section(name string) *Section {
 	return sec
 }
 
+// GetSectionWithIndex assumes named section exists and returns a new section when not.
+func (f *File) GetSectionWithIndex(name string, index int) *Section {
+	sec, err := f.GetSections(name)
```
unknwon:
```suggestion
	secs, err := f.GetSections(name)
```

file.go
```diff
@@ -171,7 +222,29 @@ func (f *File) SectionStrings() []string {
...
+	if err != nil {
+		return
+	}
+
+	for i := 0; i < len(secs); i++ {
+		// for non unique sections, always the first one is removed,
+		// so in the next iteration the following section has again index 0
+		// --> just deleting index 0 is enough
+
+		// ignore error as index 0 returns never an error
```
unknwon:
```suggestion
		// For non unique sections, always to remove the first one so in the next
		// iteration, the subsequent section continue having index 0. Ignoring the
		// error as index 0 never returns an error.
```

file.go
```diff
@@ -171,7 +222,29 @@ func (f *File) SectionStrings() []string {
...
+		_ = f.DeleteSectionWithIndex(name, 0)
+	}
+}
+
+// DeleteSectionWithIndex deletes a section.
```
unknwon:
```suggestion
// DeleteSectionWithIndex deletes a section with given name and index.
```

file.go
```diff
@@ -171,7 +222,29 @@ func (f *File) SectionStrings() []string {
...
+func (f *File) DeleteSectionWithIndex(name string, index int) error {
+	if !f.options.AllowNonUniqueSections && index != 0 {
+		return fmt.Errorf("error removing section '%s' at index %d. Index greater 0 is only allowed with non unique sections enabled", name, index)
```
unknwon:
```suggestion
		return fmt.Errorf("delete section with non-zero index is only allowed when non-unique sections is enabled", name, index)
```

file.go
```diff
@@ -181,13 +254,36 @@ func (f *File) DeleteSection(name string) {
 		name = DefaultSection
 	}
 
-	for i, s := range f.sectionList {
+	// count occurrences of the sections
```
unknwon: All comments starts with upper case, same as other places.

unknwon:
You may have missed two hidden comments in the previous round.

file.go
aligator: didn't know that, thank you

file.go
aligator: Your error string removes the %s and %d. Should I remove the name and index parameters?

aligator:
Good point, yes I missed them^^

file.go
unknwon: Yeah, I think removing these two is fine, they're passed as method arguments so the caller knows these info.

unknwon:
Please re-request review when ready :)

unknwon:
Thanks, I'm going to merge as-is and do minor changes after that :)

unknwon:
I made some changes (e.g. method rename) in https://github.com/go-ini/ini/commit/efe974c3810283243f44bc43fc50672cb203ae2f, PTAL :)

unknwon:
Also tagged v1.55.0.

aligator:
Looks good :-) 
Thank you for merging.

Fmstrat:
I was just reporting this bug myself. Nice to see a PR. When will this hit the official `gopkg.in/ini.v1`?

unknwon:
@Fmstrat https://github.com/go-ini/ini/pull/220#issuecomment-599206908

Fmstrat:
@unknwon Thanks for this. As a note (and per my other comment) it was difficult to locate how to use this (via source), and that it existed, I almost switched libraries. Could we get documentation added to the site? Not sure how to contribute there.

unknwon:
@Fmstrat yeah, I happen to be lazy for adding docs for this, but if we're going to add, there is the place: https://github.com/go-ini/docs/blob/master/en-US/howto/work_with_sections.md

PR #226, opened 2020-03-07
file: customize output delimiter

aligator:
See https://github.com/go-ini/ini/pull/203

I just added a test for the closed PR and also fixed a bug (the default delimiter was set to '' it should be '=')
I also renamed the configuration to KeyValueDelimiterOnWrite as you suggested.


unknwon:
Thanks!

PR #233, opened 2020-03-30
refactor: optimize repetitive parsing code

hellolintong:
I find parse string code is repeated in many function,  it is hard to modify, so I put it into one function.

original test cases is enough


unknwon:
Nice, thank you!

PR #235, opened 2020-04-08
add go.mod

tgulacsi:
Make it compile with Go 1.14.1

No, just compile it with Go 1.14.1


unknwon:
Thanks for the PR, but this package simply can't enable Go modules until it supports multiple module/import paths. For historical reasons, people use two different import paths for using this package.

Issue #241, opened 2020-05-20
MapTo() doesn't seem to handle embedded struct


tevino:
**Describe the bug**
`MapTo()` doesn't seem to handle embedded struct

**To Reproduce**

```go
const INI = `
[Section]
FieldInSubSection = 1
FieldInSection = 2
`

type File struct {
	Section `ini:"Section"`
}

type Section struct {
	SubSection
	FieldInSection string `ini:"FieldInSection"`
}
type SubSection struct {
	FieldInSubSection string `ini:"FieldInSubSection"`
}

func main() {
	iniFile, err := ini.Load(strings.NewReader(INI))
	if err != nil {
		log.Fatalf("load error: %v", err)
	}
	f := new(File)
	iniFile.MapTo(f)
	fmt.Printf("%+v", f)  // FieldInSubSection is not mapped
}
```

https://play.golang.org/p/cymqihBewXM

**Expected behavior**
`f.FieldInSubSection == "1"`

unknwon:
https://play.golang.org/p/7QrjK_AeYPY

```diff
type Section struct {
-	SubSection
+	SubSection     `ini:"Section"`
```

tevino:
It works!

however, is this documented somewhere? I didn't thought of doing so since "encoding/json" doesn't need the extra tag.

unknwon:
https://ini.unknwon.io/docs/advanced/map_and_reflect#other-notes-on-map%2Freflect

tevino:
@unknwon Thanks for the help so far, I tried adding the extra tag, but it doesn't work in a more complex situation. 

Here's an example of the structure I'm working with, please take a look.

I learned the usage of `nonunique` from here: https://github.com/go-ini/ini/blob/ad8a10643d24d67f464955e42b23e0d42d60fcb4/struct_test.go#L93

Am I missing something?

```go
const INI = `
[Section]
FieldInSubSection = 1
FieldInSection = 2

[Section]
FieldInSubSection = 3
FieldInSection = 4
`

type File struct {
	Sections []Section `ini:"Section,,,nonunique"`
}

type Section struct {
	SubSection     `ini:"Section"`
	FieldInSection string `ini:"FieldInSection"`
}

type SubSection struct {
	FieldInSubSection string `ini:"FieldInSubSection"`
}

func main() {
	iniFile, err := ini.LoadSources(ini.LoadOptions{AllowNonUniqueSections: true}, strings.NewReader(INI))
	if err != nil {
[...]

PR #242, opened 2020-05-25
struct: correctly mapping non unique sections to slice field

aligator:
The problem in #241

Yes, I added a test case for the case pointed out in the issue #241 by tevino.

### DIFF
--- a/struct.go
+++ b/struct.go
@@ -278,7 +278,9 @@ func parseTagOptions(tag string) (rawName string, omitEmpty bool, allowShadow bo
 	return rawName, omitEmpty, allowShadow, allowNonUnique
 }
 
-func (s *Section) mapToField(val reflect.Value, isStrict bool) error {
+// mapToField maps the given value to the matching field of the given section.
+// The sectionNumber is the index (if non unique sections are used) to which the value should be added.
+func (s *Section) mapToField(val reflect.Value, isStrict bool, sectionIndex int) error {
 	if val.Kind() == reflect.Ptr {
 		val = val.Elem()
 	}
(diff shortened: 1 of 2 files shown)


struct.go
```diff
@@ -278,7 +278,9 @@ func parseTagOptions(tag string) (rawName string, omitEmpty bool, allowShadow bo
 	return rawName, omitEmpty, allowShadow, allowNonUnique
 }
 
-func (s *Section) mapToField(val reflect.Value, isStrict bool) error {
+// mapToField maps the given value to the matching field of the given section.
+// The sectionNumber is the index (if non unique sections are used) to which the value should be added.
```
unknwon, line 282:
```suggestion
// The sectionIndex is the index (if non unique sections are enabled) to which the value should be added.
```

struct.go
```diff
@@ -307,13 +309,16 @@ func (s *Section) mapToField(val reflect.Value, isStrict bool) error {
 		}
 
 		if isAnonymous || isStruct || isStructPtr {
-			if sec, err := s.f.GetSection(fieldName); err == nil {
+			if secs, err := s.f.SectionsByName(fieldName); err == nil {
+				if len(secs) < sectionIndex {
+					return fmt.Errorf("there are not enough sections with the name %s", fieldName)
```
unknwon, line 314:
```suggestion
					return fmt.Errorf("there are not enough sections (%d <= %d) for the field %q", len(secs), sectionIndex, fieldName)
```

struct.go
```diff
@@ -307,13 +309,16 @@ func (s *Section) mapToField(val reflect.Value, isStrict bool) error {
 		}
 
 		if isAnonymous || isStruct || isStructPtr {
-			if sec, err := s.f.GetSection(fieldName); err == nil {
+			if secs, err := s.f.SectionsByName(fieldName); err == nil {
+				if len(secs) < sectionIndex {
```
unknwon, line 313:
```suggestion
				if len(secs) <= sectionIndex {
```

Off by 1?

unknwon:
New release tagged: https://github.com/go-ini/ini/releases/tag/v1.57.0

PR #253, opened 2020-08-01
Add options to control sections insensitive or keys insensitive respectively

### LINKED ISSUE #238: AWS config ini quirks
The [AWS ini config](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html) has a few quirks that are hard to handle with this library

1. The default section is `[default]`, not `[DEFAULT]` as go-ini defaults to. This is an issue when using `ini.Insensitive = true`, as the real default section name also reports as `default`. Could there be some way to override the default section when using `ini.DefaultHeader = true`?
2. Key names are case insensitive, and section names are case sensitive. go-ini's `Insensitive` handles both keys and sections with the one setting. Could there be a way to separate this?

ygj6:
fix #238 
Add options to control sections insensitive or keys insensitive respectively.

Yes,here are some test cases

### DIFF
--- a/file_test.go
+++ b/file_test.go
@@ -332,6 +332,15 @@ func TestFile_DeleteSection(t *testing.T) {
 		f.DeleteSection("")
 		So(f.SectionStrings(), ShouldResemble, []string{"author", "package"})
 	})
+	Convey("Delete a section with InsensitiveSections", t, func() {
+		f := ini.Empty(ini.LoadOptions{InsensitiveSections: true})
+		So(f, ShouldNotBeNil)
+
+		_ = f.NewSections("author", "package", "features")
+		f.DeleteSection("FEATURES")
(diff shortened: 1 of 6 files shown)


unknwon:
Thanks for the PR! Would you mind fixing the conflicts?

file_test.go
```diff
@@ -351,6 +351,15 @@ key1 = value1
 
 `)
 	})
+	Convey("Delete a section with InsensitiveSections", t, func() {
```
unknwon, line 354:
```suggestion

	Convey("Delete a section with InsensitiveSections", t, func() {
```

unknwon:
[v1.59.0](https://github.com/go-ini/ini/releases/tag/v1.59.0) has been tagged for this merge.

PR #255, opened 2020-08-09
Allow alternative delimiter for child sections

irremotus:
No problem exists.  This feature is useful to me, and maybe to others:  allowing to specify an alternative delimiter (instead of ".") for separating child section name components.

ini: added option for ChildSectionDelimiter
file: added default for ChildSectionDelimiter (".") to maintain current behavior by default
section:
  - append ChildSectionDelimiter instead of "." when checking if each section name begins with prefix
  - use ChildSectionDelimiter instead of "." when checking for LastIndex of delimiter

Yes, the new feature is tested.


unknwon:
[v1.60.0](https://github.com/go-ini/ini/releases/tag/v1.60.0) has been tagged for this merge.

PR #256, opened 2020-08-10
file: fix deleting the default section makes the next section inaccessible

### LINKED ISSUE #252: When the default section is deleted, other sections become the default section
**Describe the bug**
When DefaultHeader keeps the default setting false, I use `DeleteSection("")` to clear the default section and save the ini file, which will cause section1 to be a default section and fail to get value by  `Section("section1").Key("key1"). Value()` .
**To Reproduce**
```go
func DefaultHeader_Test(){
	cfg := ini.Empty(ini.LoadOptions{})

        cfg.Section("").Key("foo").SetValue("bar")
	cfg.Section("section1").Key("key1").SetValue("value1")
	cfg.Section("section2").Key("key2").SetValue("value2")
	cfg.DeleteSection("")
	var buf bytes.Buffer
	_, err := cfg.WriteTo(&buf)
	if err != nil {
		fmt.Println(err)
	}
	fmt.Println(buf.String())
	cfg.SaveTo("test.ini")
	cfg1,_ := ini.Load("test.ini")
	fmt.Printf("section1.key1.value1 except to be value1, actual is :%s\n", cfg1.Section("section1").Key("key1").Value())
}

func main() {
	ini.DefaultHeader = false
	DefaultHeader_Test()
}
```
```
key1 = value1

[section2]
key2 = value2


section1.key1.value1 except to be value1, actual is :
```

**Expected behavior**
```
[section1]
key1 = value1

[section2]
key2 = value2


section1.key1.value1 except to be value1, actual is :value1
```

ygj6:
fix  #252
Fix deleting the default section makes the next section inaccessible.

Yes,here are some test cases.

### DIFF
--- a/file.go
+++ b/file.go
@@ -347,7 +347,7 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
 			}
 		}
 
-		if i > 0 || DefaultHeader {
+		if i > 0 || DefaultHeader || (i == 0 && strings.ToUpper(sec.name) != DefaultSection) {
 			if _, err := buf.WriteString("[" + sname + "]" + LineBreak); err != nil {
 				return nil, err
 			}
--- a/file_test.go
+++ b/file_test.go
@@ -332,6 +332,25 @@ func TestFile_DeleteSection(t *testing.T) {
 		f.DeleteSection("")
 		So(f.SectionStrings(), ShouldResemble, []string{"author", "package"})
 	})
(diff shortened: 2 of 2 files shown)


unknwon:
Thank you!

unknwon:
[v1.58.0](https://github.com/go-ini/ini/releases/tag/v1.58.0) has tagged for this merge.

Issue #180, opened 2019-01-16
Short-circuit load


optimusgrue:
This is a feature request, not a bug.

It would possibly be useful to be able to do something like this:

`
cfg, err := ini.LazyLoad("./foo.ini", "/usr/local/etc/foo.ini", "/etc/foo.ini")
`

Instead of overwriting keys or creating shadow keys, LazyLoad would stop after the first successful loading of an ini file.

This could be done two ways:

1. Passing a new LoadOptions field named something like AllowShortCircuit to LoadSources.

2. Repeated calling LoadSources from LazyLoad.

LazyLoad is perhaps not the best name, but I am not very creative with names.

I might work on this myself, but it will be a couple of weeks before I have the time.

Edit to add: Obviously this can be done already with nested if statements, or a helper function.

optimusgrue:
A naive implementation of method 2 from above might look like this:

`
func LazyLoad(sources ...interface{}) (*File, error) {

    var fp *File
    var err error
    for i := range sources {
        fp, err = LoadSources(ini.LoadOptions{}, sources[i])
    	if err == nil {
			return fp, err
		}
	}
    return fp, err
}
`

Unknwon:
Thanks for the feedback!

>Passing a new LoadOptions field named something like AllowShortCircuit to LoadSources.

This idea looks better in my point of view.

glefloch:
@Unknwon would you accept a PR for this feature request?

Unknwon:
@glefloch Yes, a PR would be great to have!

ygj6:
This is really useful for my current project, and I made a simple change to [Reload()](https://github.com/ygj6/ini/commit/93cee51a369dde34896c3c639aad5e33585368cc) to implement this feature.
```go
func (f *File) Reload() (err error) {
	for i, s := range f.dataSources {
		if err = f.reload(s); err != nil {
			// Notice: if AllowShortCircuit and Loose are both true, AllowShortCircuit has priority over Loose.
			// Return err when all configuration sources fail to load.
			if f.options.AllowShortCircuit {
				if i < len(f.dataSources)-1 {
					continue
				}
				return err
			}
			// In loose mode, we create an empty default section for nonexistent files.
			if os.IsNotExist(err) && f.options.Loose {
				_ = f.parse(bytes.NewBuffer(nil))
				continue
			}
			return err
		}
		if f.options.AllowShortCircuit {
			return nil
		}
	}
	return nil
}
```
[...]

PR #261, opened 2020-09-01
Support Short-circuit load

ygj6:
Short-circuit load #180
1. The ShortCircuit option is adde
2. A judgment is added to the Reload() method to implement short-circuit load.

Note: When ShortCircuit and Loose are both true, ShortCircuit has priority over Loose.
Yes, all CI pass the test.

### DIFF
--- a/file.go
+++ b/file.go
@@ -302,6 +302,9 @@ func (f *File) Reload() (err error) {
 			}
 			return err
 		}
+		if f.options.ShortCircuit {
+			return nil
+		}
 	}
 	return nil
 }
--- a/ini.go
+++ b/ini.go
@@ -81,6 +81,8 @@ type LoadOptions struct {
 	IgnoreInlineComment bool
 	// SkipUnrecognizableLines indicates whether to skip unrecognizable lines that do not conform to key/value pairs.
 	SkipUnrecognizableLines bool
+	// ShortCircuit indicates whether to ignore other configuration sources after loaded the first available configuration source.
+	ShortCircuit bool
 	// AllowBooleanKeys indicates whether to allow boolean type keys or treat as value is missing.
 	// This type of keys are mostly used in my.cnf.
 	AllowBooleanKeys bool
--- a/ini_test.go
+++ b/ini_test.go
@@ -1368,7 +1368,7 @@ GITHUB = U;n;k;n;w;o;n
 			})
 		})
 
-		Convey("with `ShortCircuit` 'ture'", func() {
+		Convey("ShortCircuit", func() {
 			Convey("Load the first available configuration, ignore other configuration", func() {
 				f, err := ini.LoadSources(ini.LoadOptions{ShortCircuit: true}, minimalConf, []byte(`key1 = value1`))
 				So(f, ShouldNotBeNil)
@@ -1401,7 +1401,7 @@ E-MAIL = u@gogs.io
 `)
 			})
 
-			Convey("Inverse case", func() {
+			Convey("Ensure all sources are loaded without ShortCircuit", func() {
 				f, err := ini.LoadSources(ini.LoadOptions{ShortCircuit: false}, minimalConf, []byte(`key1 = value1`))
 				So(f, ShouldNotBeNil)
 				So(err, ShouldBeNil)
--- a/ini_test.go
+++ b/ini_test.go
@@ -1367,6 +1367,55 @@ GITHUB = U;n;k;n;w;o;n
 				}
 			})
 		})
+
+		Convey("with `ShortCircuit` 'ture'", func() {
+			Convey("Load the first available configuration, ignore other configuration", func() {
+				f, err := ini.LoadSources(ini.LoadOptions{ShortCircuit: true}, minimalConf, []byte(`key1 = value1`))
+				So(f, ShouldNotBeNil)
+				So(err, ShouldBeNil)
+				var buf bytes.Buffer
+				_, err = f.WriteTo(&buf)
+				So(err, ShouldBeNil)
+				So(buf.String(), ShouldEqual, `[author]
+E-MAIL = u@gogs.io
+
+`)
+			})
+
+			Convey("Return an error when fail to load", func() {
+				f, err := ini.LoadSources(ini.LoadOptions{ShortCircuit: true}, notFoundConf, minimalConf)
+				So(f, ShouldBeNil)
+				So(err, ShouldNotBeNil)
+			})
+
+			Convey("Used with Loose to ignore errors that the file does not exist", func() {
+				f, err := ini.LoadSources(ini.LoadOptions{ShortCircuit: true, Loose: true}, notFoundConf, minimalConf)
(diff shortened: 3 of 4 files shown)


file.go
```diff
@@ -293,15 +293,26 @@ func (f *File) reload(s dataSource) error {
 
 // Reload reloads and parses all data sources.
 func (f *File) Reload() (err error) {
-	for _, s := range f.dataSources {
+	for i, s := range f.dataSources {
 		if err = f.reload(s); err != nil {
+			// Notice: if AllowShortCircuit and Loose are both true, AllowShortCircuit has priority over Loose.
```
unknwon, line 298:
```suggestion
			// NOTE: When AllowShortCircuit and Loose are both true, AllowShortCircuit has priority over Loose.
```

file.go
```diff
@@ -293,15 +293,26 @@ func (f *File) reload(s dataSource) error {
...
+			// Return err when all configuration sources fail to load.
+			if f.options.AllowShortCircuit {
+				if i < len(f.dataSources)-1 {
+					continue
+				}
+				return err
```
unknwon, line 304: Why do we return an error here instead of relying on `if os.IsNotExist(err) && f.options.Loose {`?

ini.go
```diff
@@ -123,6 +123,9 @@ type LoadOptions struct {
 	ReaderBufferSize int
 	// AllowNonUniqueSections indicates whether to allow sections with the same name multiple times.
 	AllowNonUniqueSections bool
+	// AllowShortCircuit allows to ignore other configuration sources after loading the first available configuration source.
+	// Notice: if AllowShortCircuit and Loose are both true, AllowShortCircuit has priority over Loose.
+	AllowShortCircuit bool
```
unknwon, line 128: I think for convention, these new lines should be inserted after line 83, without prefix `Allow` because it is only used at "load" time not for key and/or values.

unknwon (changes requested):
Thanks! Please see inline comments.

file.go
ygj6: Yeah, you're right, relying on `if os.IsNotExist(err) && f.options.Loose` seems more reasonable.

ini.go
ygj6: All right, I got it. I'll submit it together later.

file.go
```diff
@@ -293,15 +293,25 @@ func (f *File) reload(s dataSource) error {
...
 			// In loose mode, we create an empty default section for nonexistent files.
 			if os.IsNotExist(err) && f.options.Loose {
 				_ = f.parse(bytes.NewBuffer(nil))
 				continue
 			}
 			return err
 		}
+		if f.options.ShortCircuit {
+			return nil
+		}
```
unknwon, line 314: I think these lines no longer useful?

ini.go
```diff
@@ -81,6 +81,9 @@ type LoadOptions struct {
 	IgnoreInlineComment bool
 	// SkipUnrecognizableLines indicates whether to skip unrecognizable lines that do not conform to key/value pairs.
 	SkipUnrecognizableLines bool
+	// ShortCircuit indicates whether to ignore other configuration sources after loading the first available configuration source.
+	// Notice: if ShortCircuit and Loose are both true, ShortCircuit has priority over Loose.
```
unknwon, line 85:
```suggestion
	// NOTE: If ShortCircuit and Loose are both true, ShortCircuit has priority over Loose.
```

file.go
```diff
@@ -293,15 +293,25 @@ func (f *File) reload(s dataSource) error {
 
 // Reload reloads and parses all data sources.
 func (f *File) Reload() (err error) {
-	for _, s := range f.dataSources {
+	for i, s := range f.dataSources {
 		if err = f.reload(s); err != nil {
+			// Notice: if ShortCircuit and Loose are both true, ShortCircuit has priority over Loose.
+			// Return err when all configuration sources fail to load.
```
unknwon, line 299:
```suggestion
```

These lines seems no longer relevant:
1. It is already obvious ShortCircuit takes priority here.
2. No longer returns an error for it.

unknwon (commented):
Thanks for the follow up! Few more styling comments.

file.go
ygj6: I think it is useful. When the two loaded configuration sources are valid, a null message is returned immediately after the first valid configuration source is loaded, and the loading of the second configuration source is stopped.
If this code is removed, both configuration sources are loaded.

unknwon: Ah, right right.

file.go
```diff
@@ -293,15 +293,25 @@ func (f *File) reload(s dataSource) error {
...
+			if f.options.ShortCircuit {
+				if i < len(f.dataSources)-1 {
+					continue
+				}
```
unknwon, line 301: Could you elaborate why this check is required?

ygj6: When all configuration sources fail to be loaded, the error information about the last configuration source is returned.
In other words, no need to `continue` when loading the last configuration source.

unknwon: >the error information about the last configuration source is returned.

Why is that a problem?

unknwon: I mean, shouldn't we return error when there is an error?

ygj6: In my opinion, in Short Circuit mode, when one of multiple configuration sources can be successfully loaded, nil should be returned. When all configuration sources fail to be loaded, the last error message should be returned.

Is there any deviation in my understanding?

unknwon: I think ShortCircuit means that it returns when there is one successfully file loaded, it is not equivalent to swallow errors.

However, to make it really useful in practice, we need to combine the behavior of Loose, i.e. ignore non-existent files.

So I think ShortCircuit and Loose are not conflict to each other but complementary. Does this make sense to you?

ygj6: I see. I agree with you. 
I'll update the PR later.

ini_test.go
```diff
@@ -1367,6 +1367,55 @@ GITHUB = U;n;k;n;w;o;n
...
+				var buf bytes.Buffer
+				_, err = f.WriteTo(&buf)
+				So(err, ShouldBeNil)
+				So(buf.String(), ShouldEqual, `[author]
+E-MAIL = u@gogs.io
+
+`)
+			})
+
+			Convey("Inverse case", func() {
```
unknwon, line 1404:
```suggestion
			Convey("Ensure all sources are loaded without ShortCircuit", func() {
```

ini_test.go
```diff
@@ -1367,6 +1367,55 @@ GITHUB = U;n;k;n;w;o;n
 				}
 			})
 		})
+
+		Convey("with `ShortCircuit` 'ture'", func() {
```
unknwon, line 1371:
```suggestion
		Convey("ShortCircuit", func() {
```

unknwon (approved):
Thank you for keeping this up!

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.61.0 has tagged for this merge.

Issue #264, opened 2020-09-15
Mapping Struct Slices Fields


kataras:
Hello @unknwon,

I am thinking to add support for loading from `.ini` for [Iris Configuration](https://github.com/kataras/iris) and custom configurations. So far we have support for json, yaml and toml and they're working fine. I have a problem though, while trying to read a config file to the `iris.Configuration` structure, I have defined the `ini` fields, I tried `allowShadow` with custom `ini.LoaderOptions` but that doesn't work either. Code speaks by itself:

```go
type Configuration struct {
 Tunneling TunnelingConfiguration `ini:"tunneling"`
}

type TunnelingConfiguration struct {
  WebInterface string `ini:"web_interface"`
  Tunnels []Tunnel `ini:"tunnels"`
}

type Tunnel struct {
  Name string `ini:"name"`
  Addr  `ini:"addr"`
}
```

I tried plenty of ini formats but I would love to support something like that (if already exists, I couldn't find it):

```ini
[tunneling]
web_interface = http://127.0.0.1:5050
[tunneling.tunnels]
name = tunnel1
addr = test1
[tunneling.tunnels]
name = tunnel2
addr = test2
```

**How I load**

```go
b, err := ioutil.ReadFile(filename)
f, err := ini.LoadSources(ini.LoadOptions{
	Insensitive:         true,
	InsensitiveKeys:     true,
	InsensitiveSections: true,
	AllowNonUniqueSections: true,
	//	AllowShadows:           true,
	DebugFunc: func(s string) {
		fmt.Printf("debug: %s\n", s)
	},
}, b)

return f.StrictMapTo(dest) // where dest is *Configuration
```

So even if `AllowNonUniqueSections` is true, the `Tunnels` are never binded to the `dest` one.

I did manage to do it by using this code,befoer StrictMapTo:

```go
	if sections, err := f.SectionsByName("tunneling.tunnels"); err == nil {
		for _, section := range sections {
			nameKey, err := section.GetKey("name")
			if err != nil || nameKey == nil {
				continue
			}
			name := nameKey.Value()
			if name == "" {
				continue
			}

			addrKey, err := section.GetKey("addr")
			if err != nil || addrKey == nil {
				continue
			}
			addr := addrKey.Value()

			dest.Tunneling.Tunnels = append(dest.Tunneling.Tunnels, iris.Tunnel{
				Name: name,
				Addr: addr,
			})
		}

	}
```

Is there a way to do that mapping automatically or it's a planned feature?

[...]

PR #265, opened 2020-09-15
Feature: add AliasMapper for sections

kataras:
Read more at: https://github.com/go-ini/ini/issues/264#issuecomment-692695551

### DIFF
--- a/file.go
+++ b/file.go
@@ -44,6 +44,7 @@ type File struct {
 	sections map[string][]*Section
 
 	NameMapper
+	AliasMapper
 	ValueMapper
 }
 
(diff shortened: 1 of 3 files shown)


file.go
```diff
@@ -156,6 +157,16 @@ func (f *File) SectionsByName(name string) ([]*Section, error) {
...
+	if f.AliasMapper != nil {
+		// Search if an alias is available instead and return as soon as possible.
+		possibleNames := f.AliasMapper(name)
+		for _, name := range possibleNames {
+			if secs := f.sections[name]; len(secs) > 0 {
+				return secs, nil
+			}
+		}
+	}
+	// Continue with the original name.
```
unknwon, line 169:
```suggestion

	// Continue with the original name.
```

section_test.go
```diff
@@ -313,3 +314,36 @@ func TestSection_DeleteKey(t *testing.T) {
 		So(f.Section("").HasKey("NAME"), ShouldBeFalse)
 	})
 }
+
+func TestSection_Alias(t *testing.T) {
+	Convey("Map to section alias", t, func() {
+
+		var sectionValueTests = map[string]string{
+			"section":        "a value",
+			"custom.section": "a value",
```
unknwon, line 323:
```suggestion
			"section":        "value a",
			"custom.section": "value b",
```

I think the test is more useful when two keys have different values. WDTY?

section_test.go
```diff
@@ -313,3 +314,36 @@ func TestSection_DeleteKey(t *testing.T) {
...
+			var cfg = struct {
+				Section struct {
+					Value string
+				}
+			}{}
+			err = f.MapTo(&cfg)
+			So(err, ShouldBeNil)
+			So(cfg.Section.Value, ShouldEqual, value)
+		}
+
```
unknwon, line 347:
```suggestion
```

unknwon (changes requested):
Thanks for the PR!

PR #266, opened 2020-09-18
Fix typo in a comment

mswdevsla:
Fix typo in a comment

Do not need to add test


unknwon (approved):
Thank you!

PR #267, opened 2020-09-19
ci: only upload coverage once per CI

ygj6:
We have multiple CI environments. It is recommended that one of them be used as the coverage statistics standard.
No test case is required.


unknwon:
Thanks for the PR! But Codecov supports uploading multiple reports for a single commit, which I think is good.

ygj6:
Thank you for your reply, I have reviewed the codecov document and will close this pr.

PR #269, opened 2020-09-24
Add extends functionality to allow embedded fields

zeripath:

Allow embedded structs to inherit from the same ini section.

```go
type Base struct {
  Inherited bool
}

type Extended struct {
  Base `ini:",,,,extends"`
  Extended bool
}
```

```ini
[section]
Inherited=true
Extended=true
```

Yes

Signed-off-by: Andrew Thornton <art27@cantab.net>

### DIFF
--- a/struct.go
+++ b/struct.go
@@ -312,7 +312,7 @@ func (s *Section) mapToField(val reflect.Value, isStrict bool, sectionIndex int,
 			fieldSection := s
 			if rawName != "" {
 				sectionName = s.name + s.f.options.ChildSectionDelimiter + rawName
-				if secs, err := s.f.SectionsByName(s.name + s.f.options.ChildSectionDelimiter + rawName); err == nil && len(secs) > 0 && sectionIndex < len(secs) {
+				if secs, err := s.f.SectionsByName(sectionName); err == nil && sectionIndex < len(secs) {
 					fieldSection = secs[sectionIndex]
 				}
 			}
--- a/struct.go
+++ b/struct.go
@@ -266,24 +266,18 @@ func setWithProperType(t reflect.Type, key *Key, field reflect.Value, delim stri
 func parseTagOptions(tag string) (rawName string, omitEmpty bool, allowShadow bool, allowNonUnique bool, extends bool) {
 	opts := strings.SplitN(tag, ",", 5)
 	rawName = opts[0]
-	if len(opts) > 1 {
-		omitEmpty = opts[1] == "omitempty"
-	}
-	if len(opts) > 2 {
-		allowShadow = opts[2] == "allowshadow"
-	}
-	if len(opts) > 3 {
-		allowNonUnique = opts[3] == "nonunique"
-	}
-	if len(opts) > 4 {
-		extends = opts[4] == "extends"
+	for _, opt := range opts[1:] {
+		omitEmpty = omitEmpty || (opt == "omitempty")
+		allowShadow = allowShadow || (opt == "allowshadow")
+		allowNonUnique = allowNonUnique || (opt == "nonunique")
+		extends = extends || (opt == "extends")
 	}
 	return rawName, omitEmpty, allowShadow, allowNonUnique, extends
 }
 
 // mapToField maps the given value to the matching field of the given section.
 // The sectionIndex is the index (if non unique sections are enabled) to which the value should be added.
-func (s *Section) mapToField(val reflect.Value, isStrict bool, sectionIndex int) error {
+func (s *Section) mapToField(val reflect.Value, isStrict bool, sectionIndex int, sectionName string) error {
 	if val.Kind() == reflect.Ptr {
 		val = val.Elem()
 	}
@@ -315,7 +309,14 @@ func (s *Section) mapToField(val reflect.Value, isStrict bool, sectionIndex int)
 			if isStructPtr && field.IsNil() {
 				field.Set(reflect.New(tpField.Type.Elem()))
 			}
-			if err := s.mapToField(field, isStrict, sectionIndex); err != nil {
+			fieldSection := s
+			if rawName != "" {
+				sectionName = s.name + s.f.options.ChildSectionDelimiter + rawName
+				if secs, err := s.f.SectionsByName(s.name + s.f.options.ChildSectionDelimiter + rawName); err == nil && len(secs) > 0 && sectionIndex < len(secs) {
+					fieldSection = secs[sectionIndex]
+				}
+			}
+			if err := fieldSection.mapToField(field, isStrict, sectionIndex, sectionName); err != nil {
 				return fmt.Errorf("map to field %q: %v", fieldName, err)
 			}
 		} else if isAnonymousPtr || isStruct || isStructPtr {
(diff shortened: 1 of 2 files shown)


lafriks (approved):

unknwon:
Thanks for the PR!

I feel like we're abusing the struct tags 😂 

Any reason why the same purpose can't be achieved by:

```go
type testExtend struct {
	BaseStruct `ini:"extended"`
	Extend     bool
}
```

Which indicates the section name directly?

zeripath:
Let's say we have the following sections:

```ini
[foo]
Base=foobase
Extended=fooextended

[Base]
Base=barbase

[fubar]
Base=fubarbase
Extended=fubarextended
```

and we want to map both `[foo]` and `[fubar]` to an `Extended{}` but a `[base]` to `Base{}`.

unknwon:
> Let's say we have the following sections:
> 
> ```ini
> [foo]
> Base=foobase
> Extended=fooextended
> 
> [Base]
> Base=barbase
> 
> [fubar]
> Base=fubarbase
> Extended=fubarextended
> ```
> 
> and we want to map both `[foo]` and `[fubar]` to an `Extended{}` but a `[base]` to `Base{}`.

@zeripath Hmm... I'm kinda confused. Could you give more concrete example input and output based on this example INI?

zeripath:
We have the following:

```ini
[cron.check_repo_stats]
Enabled=true
RunAtStart=true
Schedule=@every 24 hours

[cron.archive_cleanup]
Enabled=true
RunAtStart=true
Schedule=@every 24h
OlderThan=24h

[cron.deleted_branches_cleanup]
Enabled=true
RunAtStart=true
Schedule=@every 24h
OlderThan=24h

[cron.sync_external_users]
Enabled=true
RunAtStart=false
Schedule=@every 24h
UpdateExisting=true
```

Now it would be nice if there was a clean struct family:

```go
// BaseConfig represents the basic config for a Cron task
type BaseConfig struct {
	Enabled         bool
	RunAtStart      bool
	Schedule        string
	NoSuccessNotice bool
}

// OlderThanConfig represents a cron task with OlderThan setting
type OlderThanConfig struct {
	BaseConfig
	OlderThan time.Duration
}

// UpdateExistingConfig represents a cron task with UpdateExisting setting
type UpdateExistingConfig struct {
	BaseConfig
	UpdateExisting bool
}
```

I have code like: 

```go
	RegisterTask("check_repo_stats", &BaseConfig{
		Enabled:    true,
		RunAtStart: true,
		Schedule:   "@every 24h",
	}, func(ctx context.Context, _ *models.User, _ Config) error {
		return models.CheckRepoStats(ctx)
	})
```

where `RegisterTask` does something like:

```go
	if err := Cfg.Section("cron." + name).MapTo(config); err != nil {
		return config, err
	}
```

I therefore can't just use `ini:"extension"` or the like. I genuinely need and want the the same Section to be mapped onto the embedded-struct.

unknwon:
@zeripath Thanks for the example!

I agree with adding this this functionality, the concerns I have are:

1. We kinda having too many struct tags already... 
2. There is a conflict/undefined state when one specifies both the section name and indicates `extends`, e.g. 

	```go
	Base `ini:"section_name,,,,extends"`
	```

	Now, which one you pick? "section_name" or "extends"?

### Here is my alternative proposal

Since mapping/reflecting of a section only happens with a `struct`, we could check if a struct implements a specific interface to decide if it "extends" to the same section. 

In this package, we add an interface:

```go
// SectionExtender indicates the struct extends to the same section when mapping and reflecting.
type SectionExtender interface {
	INIExtendsSection()
}
```

Then modifying your example:

```go
// BaseConfig represents the basic config for a Cron task
type BaseConfig struct {
	ini.SectionExtender `ini:"-"`
	Enabled         bool
	RunAtStart      bool
	Schedule        string
	NoSuccessNotice bool
}

// OlderThanConfig represents a cron task with OlderThan setting
type OlderThanConfig struct {
	BaseConfig
	OlderThan time.Duration
}

// UpdateExistingConfig represents a cron task with UpdateExisting setting
type UpdateExistingConfig struct {
	BaseConfig
	UpdateExisting bool
}
```

**NOTE:** This might have slight changes during implementation.

WDYT?

zeripath:
In terms of how you'd handle the `ini:"name,,,,extends"` I think that should just use a named child section of the current section instead.

I've had a few thoughts about the `ini:",,,,extends"` bit - I mean the code handling of the commas is a bit messy and everything past the "," should just be considered a flag - much like `encoding/json` and how `xorm` handles things.

Happy to code that up and change this PR to match that

---

I don't think labelling the base struct with an interface is a good idea - the struct doing the embedding should have control over how it's embedded structs are mapped. You may not always be able to or want to alter the base struct. It's really the wrong place to do this kind of thing.

unknwon:
>
```diff
+	for _, opt := range opts[1:] {
+		omitEmpty = omitEmpty || (opt == "omitempty")
+		allowShadow = allowShadow || (opt == "allowshadow")
+		allowNonUnique = allowNonUnique || (opt == "nonunique")
+		extends = extends || (opt == "extends")
```

Brilliant! Thanks for the patch ❤️ 

Alright, agree with the approach implemented in this PR then.

> In terms of how you'd handle the `ini:"name,,,,extends"` I think that should just use a named child section of the current section instead.

As long as this is also addressed/implemented.

zeripath:
patch applied.

struct.go
```diff
@@ -295,20 +292,34 @@ func (s *Section) mapToField(val reflect.Value, isStrict bool, sectionIndex int)
...
 
-		if isAnonymous || isStruct || isStructPtr {
+		if extends && (isAnonymousPtr || (isStruct && tpField.Anonymous)) {
+			if isStructPtr && field.IsNil() {
+				field.Set(reflect.New(tpField.Type.Elem()))
+			}
+			fieldSection := s
+			if rawName != "" {
+				sectionName = s.name + s.f.options.ChildSectionDelimiter + rawName
+				if secs, err := s.f.SectionsByName(s.name + s.f.options.ChildSectionDelimiter + rawName); err == nil && len(secs) > 0 && sectionIndex < len(secs) {
```
unknwon, line 315:
```suggestion
				if secs, err := s.f.SectionsByName(sectionName); err == nil && sectionIndex < len(secs) {
```

unknwon: I think `len(secs) > 0` is covered when `sectionIndex < len(secs)` is true. 

In the case of a bad `sectionIndex` (i.e. -1), [this value is not controlled by user input](https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/go-ini/ini%24%4061851be+mapToField&patternType=literal) so I don't think it's a problem. At very least we can have a check to make sure `sectionIndex` is never smaller than 0 in the beginning of this method.

struct_test.go
```diff
@@ -364,6 +387,18 @@ names=alice, bruce`))
 	})
 }
 
+func Test_MapToExtended(t *testing.T) {
```
unknwon, line 390: The body of the test case seems identical to the one above, should we only keep one?

unknwon (changes requested):
Thanks for the follow up! All looking good except few comments.

struct_test.go
zeripath: Yes sorry this sneaked as I was testing this code whilst I was making changes to this code.

unknwon: No worries!

struct_test.go
```diff
@@ -364,6 +387,18 @@ names=alice, bruce`))
...
+		f, err := ini.Load([]byte(confDataStruct))
+		So(err, ShouldBeNil)
+		So(f, ShouldNotBeNil)
+		te := testExtend{}
+		So(f.Section("extended").MapTo(&te), ShouldBeNil)
+		So(te.Base, ShouldBeTrue)
+		So(te.Extend, ShouldBeTrue)
+	})
+}
+
```
unknwon, line 401:
```suggestion
```

unknwon (approved):

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.62.0 has been tagged for this merge.

zeripath:
Thanks!

PR #281, opened 2021-01-15
fix case where the user requests the empty section after removing the DEFAULT section

tphilippeau-tower:

When the DEFAULT section is removed from the ini file, it cannot be requested back in using the empty string.
It currently return a nil pointer, instead of a new section

Yes


unknwon:
Hey! This PR was accidentally closed because `master` branch is no longer exists, would you mind re-open one against the new `main` branch? Thank you!

tphilippeau-tower:
Done

PR #285, opened 2021-03-12
Add AllowDuplicateShadowValues option

### LINKED ISSUE #284: Cannot Read Duplicate Values for Shadowed Keys
**Describe the bug**
Duplicate values for shadowed keys are not preserved in either `ValueWithShadows()` or just `WriteTo()`.

**To Reproduce**
```go
package main

import (
	"os"

	"gopkg.in/ini.v1"
)

const f = `[test]
Key1=Key1Value1
Key1=Key1Value2
Key1=Key1Value1
`

func main() {
	cfg, _ := ini.ShadowLoad([]byte(f))
	cfg.WriteTo(os.Stdout)
}

```

produces

```ini
[test]
Key1 = Key1Value1
Key1 = Key1Value2
```

**Expected behavior**
Reading in the file should not destroy the duplicate line. `WriteTo()` should reproduce the source file:
```ini
[test]
Key1 = Key1Value1
Key1 = Key1Value2
Key1 = Key1Value1
```

**Additional context**
While similarly named, this is not a duplicate of #283 which is a case of needing `ValueWithShadows()`.

I am working with a save file for a game that is stored in an INI format. Extra equipment may be stored in the cargo with lines such as 
```ini
cargo                = 2322110156, 1, , , 0
cargo                = 2859002568, 1, , , 0
cargo                = 2859002568, 1, , , 0
```
where the last two duplicate lines indicate additional quantity. Currently, `go-ini` is mangling the duplicate lines resulting in the loss of equipment.

dwmunster:

Fixes #284.

Test cases have been added to key_test.go for both the default and new cases.

### DIFF
--- a/ini.go
+++ b/ini.go
@@ -125,6 +125,8 @@ type LoadOptions struct {
 	ReaderBufferSize int
 	// AllowNonUniqueSections indicates whether to allow sections with the same name multiple times.
 	AllowNonUniqueSections bool
+	// AllowDuplicateShadowValues indicates whether values for shadowed keys should be deduplicated.
+	AllowDuplicateShadowValues bool
 }
 
 // DebugFunc is the type of function called to log parse events.
--- a/key.go
+++ b/key.go
@@ -54,13 +54,15 @@ func (k *Key) addShadow(val string) error {
 		return errors.New("cannot add shadow to auto-increment or boolean key")
 	}
 
-	// Deduplicate shadows based on their values.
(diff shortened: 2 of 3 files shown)


dwmunster:
@unknwon moving the default branch to main seems to have closed pull requests against it. Would you like me to reopen against main?

unknwon:
Oops my bad! If you could reopen it, that would be great!

Issue #287, opened 2021-04-02
Allow for lower case [default] section


BernhardLenz:
The aws cli uses a lower case [default] notation for the default section. ini should allow an overwrite of the name of the default section, e.g. through ini.DefaultSection = "default"

**To Reproduce**
Add section [default] to ini file with some random key/value pairs
Loop over each key

	for _, fileSection := range file.Sections() {
		sectionName := fileSection.Name() //sectionName will be DEFAULT even though it is named default in the file
		for _, key := range fileSection.Keys() {//no keys available even though the [default] section has entries

Overall though the uppercase lowercase spellings in the ini file need to be preserved though, hence the file cannot be completely converted into lowercase.

PR #293, opened 2021-06-21
Making DefaultSection modifyable to handle aws profile default.

jade-clarke:
This was done to fix the handling of aws credential files. If the default section is not modifyable then when the profiles are loaded from the credentials file, an erronious non-existent DEFAULT section is returned as part of the results array. When given the ability to modify DefaultSection and set it to lower case 'default', it loads the credentials as would be expected.

### DIFF
--- a/deprecated.go
+++ b/deprecated.go
@@ -14,12 +14,10 @@
 
 package ini
 
-const (
-	// Deprecated: Use "DefaultSection" instead.
-	DEFAULT_SECTION = DefaultSection
-)
-
 var (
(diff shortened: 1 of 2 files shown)


jade-clarke:
This was done to fix this issue: https://github.com/go-ini/ini/issues/287

unknwon:
Hey! This PR was accidentally closed because `master` branch is no longer exists, would you mind re-open one against the new `main` branch? Thank you!

jade-clarke:
@unknwon I have reopened this in https://github.com/go-ini/ini/pull/319

PR #299, opened 2021-09-06
Fix parsing python multiline string

gandarez:

This PR fixes parsing Python multiline strings to keep the original indentation instead of removing all leading spaces. It breaks some compatibility with legacy Python parsers as them fail to parse multiline strings that are malformed. This issue was first reported [here](https://github.com/wakatime/wakatime-cli/issues/532).

Yes

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -302,7 +302,6 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
 	parserBufferPeekResult, _ := p.buf.Peek(bufferSize)
 	peekBuffer := bytes.NewBuffer(parserBufferPeekResult)
 
-	indentSize := 0
 	for {
 		peekData, peekErr := peekBuffer.ReadBytes('\n')
 		if peekErr != nil {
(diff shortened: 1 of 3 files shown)


gandarez:
@unknwon any update here?

unknwon:
Thanks for the PR! Will take a look when time permits.

parser.go
```diff
@@ -329,28 +328,14 @@ func (p *parser) readPythonMultilines(line string, bufferSize int) (string, erro
...
-			indentSize = currentIndentSize
-			p.debug("readPythonMultilines: indent size is %d", indentSize)
-		}
-
-		// Make sure each line is indented at least as far as first line.
-		if currentIndentSize < indentSize {
-			p.debug("readPythonMultilines: end of value, current indent: %d, expected indent: %d, line: %q", currentIndentSize, indentSize, line)
-			return line, nil
-		}
-
```
unknwon, line 344: What is the rationale of deleting these lines? Without these, it seems we can't determine indent size for the multiline value?

gandarez: The main idea here is to not bypass any miss indentation from the source which means we're looking for **break lines** and not for "malformed" multiline values. Having it in the logic makes the first line to always determine the other's indentation.

unknwon: > The main idea here is to not bypass any miss indentation from the source which means we're looking for **break lines** and not for "malformed" multiline values. Having it in the logic makes the first line to always determine the other's indentation.

What is definition of a "break line", is it a line that failed to be matched with `pythonMultiline.FindStringSubmatch`?

unknwon: All look good to me, just want to confirm the impact of deleting these lines.

gandarez:
Hi, any update on this?

bryanpedini:
> Hi, any update on this?

From the history of the comments in @unknwon's review, it seems he's waiting for some clarifications from your side on the deleted lines.

gandarez:
The detailed explanatation for this issue was first reported [here](https://github.com/wakatime/wakatime-cli/issues/532). In a nutshell it's removing the original indentation when saving.

unknwon (approved):

unknwon:
@gandarez LGTM! Could you fix the conflicts so CI can be kicked off?

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.65.0 has been tagged for this release.

PR #301, opened 2021-09-09
Fix creation of default section

tphilippeau-tower:

When the DEFAULT section is removed from the ini file, it cannot be requested back in using the empty string.
It currently return a nil pointer, instead of a new section

Yes


unknwon (approved):
Thank you!

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.64.0 has been tagged for this merge.

PR #306, opened 2021-10-29
Fix nested values can span sections

ravron:
When configured with `AllowNestedValues: true`, the parser attempts to parse values "nested" under a key, [AWS-style](https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#nested-values). To trigger this behavior, the parent key must have an empty value, and child keys must be indented. However, if the next value is in a different section, it cannot be a nested value, but the parser can treat it as such anyways. As an example:

```
[section]
key1 = value1
key2 =
[section2]
  key3 = value3
```

Before this PR, `key3` is treated as a nested value under `key2`, and `section2` has no keys. 

Yes, I've added tests that cover the issue and fail on the current tip of `main`.

### DIFF
--- a/ini_test.go
+++ b/ini_test.go
@@ -1562,3 +1562,75 @@ Eu feugiat pretium nibh ipsum consequat nisl vel pretium lectus. Habitant morbi 
 		testData.Value3,
 	)
 }
+
+func Test_NestedValuesSpanningSections(t *testing.T) {
+	t.Run("basic nested value", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


unknwon:
GitHub Action got stuck and does not run, I'm gonna try close and reopen the PR see if helps.

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.66.2 has been tagged for this merge.

Issue #271, opened 2020-11-09
`HasSection` function proposal


bryanpedini:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
`HasSection(string) (boolean, error) {}` function, to be able to use "quick if statements" without needing to check if the section name is in `cfg.Sections()`.

**Describe alternatives you've considered**
Check if section name is in the slice provided by `cfg.Sections()`.

**Additional context**
None, I think.

unknwon:
@bryanpedini Thanks for the feedback! Any reason why the method signature is not `HasSection(string) (bool)` (without `error`)?

bryanpedini:
Honestly dunno what I thought 2 weeks ago, was on a local shady proof of concept now deleted project that I can't currenlty remember anything about. 😅
It was probably about - if `cfg.Sections()` is empty or if it throws an error itself down the line of function recursiveness or something like that.

Honestly, if it works great even without the `error`, it's even better 'cause well, it doesn't throw errors...
If you're thinking of an implementation, please feel free to do it whaever way you like the most!

bryanpedini:
Hello, sorry for the ping.
Any news regarding #308 related to this FR?

PR #308, opened 2021-11-19
file: add `HasSection` method

bryanpedini:
copied implementation from section.go

Feature proposal #271.

Yes indeed (after seeing this question, but I digress).

### DIFF
--- a/file_test.go
+++ b/file_test.go
@@ -243,11 +243,11 @@ func TestFile_HasSection(t *testing.T) {
 	require.NotNil(t, f)
 
 	sec := f.HasSection("author")
-	assert.Equal(t, true, sec)
+	assert.True(t, sec)
 
 	t.Run("section not exists", func(t *testing.T) {
 		nonexistent := f.HasSection("404")
-		assert.Equal(t, false, nonexistent)
+		assert.False(t, nonexistent)
 	})
 }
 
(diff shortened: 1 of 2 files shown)


file_test.go
```diff
@@ -237,6 +237,20 @@ func TestFile_GetSection(t *testing.T) {
...
+	f, err := Load(fullConf)
+	require.NoError(t, err)
+	require.NotNil(t, f)
+
+	sec := f.HasSection("author")
+	assert.Equal(t, true, sec)
+
+	t.Run("section not exists", func(t *testing.T) {
+		nonexistent := f.HasSection("404")
+		assert.Equal(t, false, nonexistent)
```
unknwon, line 250:
```suggestion
		assert.False(t, nonexistent)
```

file_test.go
```diff
@@ -237,6 +237,20 @@ func TestFile_GetSection(t *testing.T) {
 	})
 }
 
+func TestFile_HasSection(t *testing.T) {
+	f, err := Load(fullConf)
+	require.NoError(t, err)
+	require.NotNil(t, f)
+
+	sec := f.HasSection("author")
+	assert.Equal(t, true, sec)
```
unknwon, line 246:
```suggestion
	assert.True(t, sec)
```

unknwon (approved):
Thank you!

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.66.0 has been tagged for this merge.

PR #309, opened 2021-12-01
Fix python multiline EOF error

gandarez:

When parsing `python multiline` values and the key which is being parsed is the last one on the ini file and there's no new line at the end, then it throws an error. To fix that we might not consider EOF as an error when reading the last line because it's either can be a `\n` or `EOF`.

This problem was first reported [here](https://github.com/wakatime/wakatime-cli/issues/593).

The `TestPythonMultiline_EOF` was added.

### DIFF
--- a/ini_test.go
+++ b/ini_test.go
@@ -1562,3 +1562,26 @@ func TestPythonMultiline(t *testing.T) {
 		testData.Value3,
 	)
 }
+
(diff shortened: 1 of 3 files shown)


unknwon (approved):
Thank you!

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.66.1 has been tagged for this merge.

Issue #310, opened 2021-12-05
ValueWithShadows returning a non empty array when there are no values


NDagestad:
When getting the values and shadows of a key, if there are no values at all, StringsWithShadows will return an array with an empty string. This does not seem like the right behaviour.
I would expect StringsWithShadows to return an empty array if there are no values. The documentation says that ValueWithShadows returns the raw value so there is a case to be made for it to return an empty string, but I would argue it is more intuitive to return an empty array even there.

Something like this should be enough to fix it from what I can tell (it is good enough in my case at least)
```diff
diff --git a/key.go b/key.go
index 0302c29..2147974 100644
--- a/key.go
+++ b/key.go
@@ -113,6 +113,9 @@ func (k *Key) Value() string {
 // ValueWithShadows returns raw values of key and its shadows if any.
 func (k *Key) ValueWithShadows() []string {
        if len(k.shadows) == 0 {
+               if k.value == "" {
+                       return []string{}
+               }
                return []string{k.value}
        }
        vals := make([]string, len(k.shadows)+1)
```

PR #313, opened 2022-01-12
Return an empty array from `ValueWithShadows` if there is none

NDagestad:
When there are no value for a given key, ValueWithShadows used to return an array containing an empty string. Instead, this PR makes it return an empty array. This change produces the same behaviour for StringsWithShadows as it uses ValueWithShadows to do it's job.

Link to the issue: https://github.com/go-ini/ini/issues/310

### Checklist

- [x] I agree to follow the [Code of Conduct](https://go.dev/conduct) by submitting this pull request.
- [x] I have read and acknowledge the [Contributing guide](https://github.com/go-ini/ini/blob/main/.github/contributing.md).
There aren't any existing test cases for ValueWithShadows. Should I add a new one ?

PS: I tried to wait a little before doing the PR in case any newcomer wanted to use the opportunity of this simple fix for their first PR, but it's been a month now so I'll  just do it myself ¯\\\_(ツ)_/¯

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,3 +7,6 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*_test.go]
+trim_trailing_whitespace = false
--- a/file.go
+++ b/file.go
@@ -442,30 +442,16 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
 				kname = `"""` + kname + `"""`
 			}
 
-			shadows := key.ValueWithShadows()
-			if len(shadows) == 0 {
+			writeKeyValue := func(val string) (bool, error) {
 				if _, err := buf.WriteString(kname); err != nil {
-					return nil, err
-				}
-				// Write out alignment spaces before "=" sign
-				if PrettyFormat {
-					buf.Write(alignSpaces[:alignLength-len(kname)])
-				}
-				if _, err := buf.WriteString(equalSign + LineBreak); err != nil {
-					return nil, err
-				}
-			}
-
-			for _, val := range shadows {
-				if _, err := buf.WriteString(kname); err != nil {
-					return nil, err
+					return false, err
 				}
 
 				if key.isBooleanType {
 					if kname != sec.keyList[len(sec.keyList)-1] {
 						buf.WriteString(LineBreak)
 					}
-					continue KeyList
+					return true, nil
 				}
 
 				// Write out alignment spaces before "=" sign
@@ -482,7 +468,24 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
 					val = `"` + val + `"`
 				}
 				if _, err := buf.WriteString(equalSign + val + LineBreak); err != nil {
+					return false, err
+				}
+				return false, nil
+			}
+
+			shadows := key.ValueWithShadows()
+			if len(shadows) == 0 {
+				if _, err := writeKeyValue(""); err != nil {
 					return nil, err
+				}
+			}
+
+			for _, val := range shadows {
+				exit_loop, err := writeKeyValue(val)
+				if err != nil {
+					return nil, err
+				} else if exit_loop {
+					continue KeyList
 				}
 			}
 
--- a/file.go
+++ b/file.go
@@ -481,10 +481,10 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
 			}
 
 			for _, val := range shadows {
-				exit_loop, err := writeKeyValue(val)
+				exitLoop, err := writeKeyValue(val)
 				if err != nil {
 					return nil, err
-				} else if exit_loop {
+				} else if exitLoop {
 					continue KeyList
 				}
 			}
--- a/file.go
+++ b/file.go
@@ -442,7 +442,21 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
 				kname = `"""` + kname + `"""`
 			}
 
-			for _, val := range key.ValueWithShadows() {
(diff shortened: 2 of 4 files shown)


unknwon:
> * There aren't any existing test cases for ValueWithShadows. Should I add a new one ?

Yes, please do so!

NDagestad:
It was wise of you to make me do tests, I found out that my original way of doing it broke a few things that I did not see because it had no effect on my use case. I also edited the .editorconfig to not trim trailing spaces in test files since it got me when I first saved the file after adding my test 🙃

.editorconfig
```diff
@@ -7,3 +7,6 @@ charset = utf-8
 end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
+
+[*_test.go]
+trim_trailing_whitespace = false
```
unknwon, line 12: I don't see any trailing whitespace in the `key_test.go`, how is this making a difference?

unknwon (changes requested):
Thanks for the follow up!

NDagestad:
In TestKey_NestedValues, the s3 key has a trailling space after the equal sign
in the expected output in the assertion.

NDagestad:
Ha! I forgot to give the test case a description, and it seems that it is not ran, did I miss something ? (Haven't done tests in go before so it is not unlikely I did something wrong)

NDagestad:
Third time's the charm !
This time I have checked that my test is being run, and everything works as it should.
It turns out that my seccond attempt was flawed but the first method was working as intended.
If the Key is produced by the `section.Key()` function, using `section.HasKey()` will not work since it has been
added to the section in `Section.Key()`. Simply deciding to return an empty array based on the
content of `Key.Value` does work though, but `File.WriteToBuffer()` loops on the output
`Key.ValueWithShadow()` so if an empty array is returned, that key will not be added to the output
even if it was in the input but had no value (I assume this is a behaviour that you want to keep, and I
would agree.)
So I had to add a special case to be handled before the loop in case an empty array was returned.

file.go
```diff
@@ -442,7 +442,21 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
...
+				// Write out alignment spaces before "=" sign
+				if PrettyFormat {
+					buf.Write(alignSpaces[:alignLength-len(kname)])
+				}
+				if _, err := buf.WriteString(equalSign + LineBreak); err != nil {
+					return nil, err
+				}
+			}
+
+			for _, val := range shadows {
```
unknwon, line 459: I think this block is missing many handlings from where it is copied from (below).

We should either make this loop (starts on line 459) body to be an inline function .e.g `writeKey := func(k *Key) error {...}`, or do something like:

```go
keys := make([]*Key, 1, len(shadows)+1)
keys[0] = key
keys = append(keys, shadows)
for _, val := range keys {
	...
```

But the latter is obviously more expensive IMO (due to allocations).

unknwon (changes requested):

file.go
NDagestad: I do not see what handlings are missing, the value check don't have any meaning since it is in a case where the key has no associated value, and I imagine a boolean key has a value so it wouldn't pass the condition and be handled in the loop like before, am I missing something ?
Anyways I am not a fan of the duplicated code like I did, I do prefer the first solution you proposed (partly because I don't really get how the second one would work) do you really prefer the function to be inline ? If so why ? (Not that I really have anything against it, I am just curious)

NDagestad: Also, I see codecov is not happy with my patches, is it something I should care about ? (`go test -cover` gives a different code coverage value from codecov)

unknwon: > I do not see what handlings are missing, the value check don't have any meaning since it is in a case where the key has no associated value, and I imagine a boolean key has a value so it wouldn't pass the condition and be handled in the loop like before, am I missing something ?

Ah, you're right. So alternatively, you could add a code commend explaining why we don't need to deal with values here.

But since we can factor out a function, I think having a unified logic with an inline function is relatively more robust (if anything changes to the key name handling).

>(partly because I don't really get how the second one would work)
Hmm, that was for illustration mostly, on a second look I also forgot how it should work 😅 so forgot about it.

>do you really prefer the function to be inline ? If so why ? (Not that I really have anything against it, I am just curious)

Yes, because I don't a reason why it shouldn't be inlined. It is only used here and never meant to be used outside, and the function body is also pretty sophisticated.

unknwon: >Also, I see codecov is not happy with my patches, is it something I should care about ? (go test -cover gives a different code coverage value from codecov)

Yeah... the difference is annoying, CodeCov is more FYI, don't need to worry much unless diff coverage is 0.

NDagestad:
I was a bit on the fence about passing only the value to write in the function and using `key` and `kname` from the calling function, but it would ba a bit redundant to pass them as parameters and `kname` is realistically never going to change since the function is declared for the current `key` (don't know if what I said really made any sense)

file.go
```diff
@@ -468,10 +468,27 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
...
+				if _, err := writeKeyValue(""); err != nil {
 					return nil, err
 				}
 			}
 
+			for _, val := range shadows {
+				exit_loop, err := writeKeyValue(val)
+				if err != nil {
+					return nil, err
+				} else if exit_loop {
```
unknwon, line 487:
```suggestion
				exitLoop, err := writeKeyValue(val)
				if err != nil {
					return nil, err
				} else if exitLoop {
```

unknwon (approved):

unknwon:
Thank you, merging!

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.66.3 has been released for this merge.

PR #323, opened 2022-05-30
parser: fix line skipping when key is empty

gandarez:
### Describe the pull request

This PR fixes a scenario where the key name is zero length and `readKeyName` incorrectly handle it. Even if setting `SkipUnrecognizableLines` to **true** wasn't being skipped. 

### Checklist

- [x] I agree to follow the [Code of Conduct](https://go.dev/conduct) by submitting this pull request.
- [x] I have read and acknowledge the [Contributing guide](https://github.com/go-ini/ini/blob/main/.github/contributing.md).

### DIFF
--- a/parser.go
+++ b/parser.go
@@ -161,9 +161,13 @@ func readKeyName(delimiters string, in []byte) (string, int, error) {
 	}
 
 	endIdx = strings.IndexAny(line, delimiters)
-	if endIdx <= 0 {
+	if endIdx < 0 {
 		return "", -1, ErrDelimiterNotFound{line}
 	}
+	if endIdx == 0 {
+		return "", -1, ErrEmptyKeyName{line}
+	}
+
 	return strings.TrimSpace(line[0:endIdx]), endIdx + 1, nil
 }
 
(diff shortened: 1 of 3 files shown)


parser.go
```diff
@@ -161,7 +161,7 @@ func readKeyName(delimiters string, in []byte) (string, int, error) {
 	}
 
 	endIdx = strings.IndexAny(line, delimiters)
-	if endIdx < 0 {
+	if endIdx <= 0 {
 		return "", -1, ErrDelimiterNotFound{line}
```
unknwon, line 165: I think we need to return a different error here, because `endIdx == 0` should be `ErrKeyEmptyName` not `ErrDelimiterNotFound`.

unknwon (changes requested):
Thanks for the PR!

One minor issue.

parser.go
gandarez: Makes sense to have a new error for that. I already added it to the fixup commit.

unknwon (approved):
Thanks for the followup!

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.66.6 has been tagged for this merge.

PR #331, opened 2022-11-11
add NoMultiLineSurroundingQuotes option

benbaker76:
### Describe the pull request

Consider the following multi-line ini file:
```ini
value1 = some text here
	some more text here 2
```

Run the following code:
```go
path := "multiline_eof.ini"
f, _ := ini.LoadSources(ini.LoadOptions{
  AllowPythonMultilineValues: true,
}, path)

f.Section("").Key("value1").SetValue("some text here\n\tsome more text here 2")
f.SaveTo(path)
```

Will output:
```ini
value1 = """some text here
	some more text here 2"""
```

Now run:
```go
path := "multiline_eof.ini"
f, _ := ini.LoadSources(ini.LoadOptions{
  AllowPythonMultilineValues: true,
  NoMultiLineSurroundingQuotes: true,
}, path)

f.Section("").Key("value1").SetValue("some text here\n\tsome more text here 2")
f.SaveTo(path)
```

The output is the same as the original file
```ini
value1 = some text here
	some more text here 2
```

Link to the issue: n/a

### Checklist

- [*] I agree to follow the [Code of Conduct](https://go.dev/conduct) by submitting this pull request.
- [*] I have read and acknowledge the [Contributing guide](https://github.com/go-ini/ini/blob/main/.github/contributing.md).
- [*] I have added test cases to cover the new code.

### DIFF
--- a/file.go
+++ b/file.go
@@ -460,7 +460,7 @@ func (f *File) writeToBuffer(indent string) (*bytes.Buffer, error) {
 				}
 
 				// In case key value contains "\n", "`", "\"", "#" or ";"
-				if strings.ContainsAny(val, "\n`") {
+				if !f.options.NoMultiLineSurroundingQuotes && strings.ContainsAny(val, "\n`") {
 					val = `"""` + val + `"""`
 				} else if !f.options.IgnoreInlineComment && strings.ContainsAny(val, "#;") {
 					val = "`" + val + "`"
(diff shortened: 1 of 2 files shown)


aghandri (changes requested):
Thank you for suggesting this change, it is indeed a very handy option, especially when used to generate aws config files.

However, the result of your changes still keeps the surrounding double quotes if the value contains leading and trailing white spaces.

Here is the result I obtained: 

```ini
duration_seconds    = xxxxxxx
s3                  = "
        signature_version = version"
```

To remove these double quotes, I suggest to add the condition on line 467:
```go
} else if !f.options.NoMultiLineSurroundingQuotes && len(strings.TrimSpace(val)) != len(val) {
	val = `"` + val + `"`
}
```
The result would be:
```ini
duration_seconds    = xxxxxxx
s3                  = 
        signature_version = version
```

PR #351, opened 2024-05-01
Add go.mod

tredoe:
### Changes

- Add go.mod
- Better .gitignore


bewuethr:
This seems to be addressed in #374.

PR #383, opened 2026-05-09
struct: return error instead of panic when MapTo receives a nil pointer (fixes #369)

### LINKED ISSUE #369: panic: nil struct pointer to MapTo
### Version

1.67.0

### Describe the bug

This code should gracefully fail with an error, panics instead

### To reproduce

```
package main

import (
    "log"

    "gopkg.in/ini.v1"
)

type Config struct {
    User string `ini:"user"`
    Pass string `ini:"pass"`
}

const iniData = `
user = alice
pass = secret
`

func main() {
    var cfg *Config

    f, err := ini.Load([]byte(iniData))
    if err != nil {
        log.Fatal(err)
    }

    if err := f.MapTo(&cfg); err != nil {
        log.Fatal(err)
    }

    log.Printf("User: %s, Pass: %s\n", cfg.User, cfg.Pass)
}
```

### Expected behavior

Gracefully return an error

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

### LINKED ISSUE #370: panic: slice of struct pointers
### Version

1.67.0

### Describe the bug

The library doesn't treat correctly slices of struct pointers

### To reproduce
```
package main

import (
	"log"

	"gopkg.in/ini.v1"
)

const iniData = `[Peer]
PublicKey    = pubClientKey
PresharedKey = psKey
AllowedIPs   = 10.2.0.2/32,fd00:2::2/128

[Peer]
PublicKey    = pubClientKey2
PresharedKey = psKey2
AllowedIPs   = 10.2.0.3/32,fd00:2::3/128
`

type Peer struct {
	PublicKey    string
	PresharedKey string
	AllowedIPs   []string `delim:","`
}

type Config struct {
	Peers []*Peer `ini:"Peer,nonunique"`
}

func main() {
	cfgFile, err := ini.LoadSources(ini.LoadOptions{AllowNonUniqueSections: true}, []byte(iniData))
	if err != nil {
		log.Fatal(err)
	}

	cfg := new(Config)
	if err := cfgFile.MapTo(&cfg); err != nil {
		log.Fatal(err)
	}
}
```

### Expected behavior

Should unmarshal correctly, panics instead

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

nghiack7:
## Problem

Two related bugs in `mapToField` when the target is a nil pointer or a slice of struct pointers:

- **#369** — `MapTo` panics when the destination struct field is a nil pointer (e.g. `var s *S; f.MapTo(&s)`)
- **#370** — `MapTo` panics when mapping into a `[]*Struct` field (slice of struct pointers), common in WireGuard-style INI configs with repeated `[Peer]` sections

Both root causes are the same: `mapToField` only unwrapped one level of pointer indirection and never initialized nil pointers.

## Fix

Replace the single `if val.Kind() == reflect.Pointer` block with a `for` loop that:
1. Walks the full pointer chain (handles `**T`, `***T`, etc.)
2. Allocates any nil pointer via `reflect.New` (mirrors `encoding/json` behavior)
3. Returns a clear error if the pointer cannot be set

## Tests

- `Test_MapToNilPointer` — regression for #369; verifies nil pointer is auto-initialized and fields are populated
- `Test_MapToSliceOfStructPointers` — regression for #370; verifies `[]*Peer` with `AllowNonUniqueSections` maps correctly

Fixes #369, fixes #370


nghiack7:
Friendly ping — this PR is ready for review. Fixes panics when `MapTo` receives a nil pointer or a slice of struct pointers (issues #369, #370). Tests pass ✅.

PR #385, opened 2026-06-04
Optimize `Key.Strings` by reducing memory re-allocations and simplifying control flow

gitKashish:
### Describe the pull request

`Key.Strings` was reimplemented to improve performance on large INI arrays. The original implementation converted the entire string to `[]rune` upfront on every call, used `bytes.Buffer` with no pre-sizing, and assumed a fixed slice capacity of 2 regardless of actual delimiter count. On large arrays this caused excessive heap allocations that made the function a significant bottleneck.

This was identified as the root cause of a Forgejo startup performance regression where instances with large lists in config files experienced severely degraded startup times due to repeated calls to `Key.Strings` during config parsing.

Link to the issue: https://codeberg.org/forgejo/forgejo/issues/11453

The new implementation:
- Works directly on the raw string, decoding runes lazily with `utf8.DecodeRuneInString` only at write time — avoiding the full `[]rune` allocation
- Uses `strings.Builder` pre-grown to `len(str)` to eliminate internal buffer reallocations
- Pre-sizes the result slice using `strings.Count(str, delim) + 1` so the slice is correctly sized on the first allocation

Benchmarks on a large INI array (10 runs, Windows/amd64, i5-11300H):

```
                      ns/op        ±stddev      B/op         allocs/op
NEW  Strings         379,108       ±21,950      116,384        3,854
OLD  _Strings      1,987,842       ±89,800      400,248       29,074

delta ns/op:   -80.9% (5.24x faster)
delta B/op:    -70.9% (116KB vs 390KB)
delta allocs:  -86.8% (3,854 vs 29,074)
```
[Raw `go bench` output](https://github.com/user-attachments/files/28610571/large_array_bench.txt)

### Checklist
- [x] I agree to follow the [[Code of Conduct](https://go.dev/conduct)](https://go.dev/conduct) by submitting this pull request.
- [x] I have read and acknowledge the [[Contributing guide](https://github.com/go-ini/ini/blob/main/.github/contributing.md)](https://github.com/go-ini/ini/blob/main/.github/contributing.md).


unknwon:
@gitKashish looks like the code is not compiling.

gitKashish:
Apologies for the oversight! While cleaning up the code after running the benchmarks, I accidentally removed the new `unicode/utf8` import instead of the old `bytes` import. I've just pushed a fix and verified that everything compiles cleanly with `go build`.

unknwon (approved):
Thank you!

unknwon:
https://github.com/go-ini/ini/releases/tag/v1.67.3 published

gitKashish:
Thank you very much! @unknwon

